PR #130, opened 2018-01-23
Descending Dropdown

### LINKED ISSUE #78: Version pulldown should sort in logical order
2017-12-06 at 4 42 pm

dabria:
Closes #78

Adds reverse ordering for version dropdowns in both the nav bar and drawer (when scaled down)
screenshot 2018-01-22 04 12 22

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -64,9 +64,9 @@ pygmentsUseClasses=true
         weight = 2
         latest = "2.8"
         [params.products.sensu_enterprise.versions]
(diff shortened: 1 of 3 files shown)


calebhailey:
Why does it look gnarly? I think this is great!

cwjohnston:
@calebhailey I think the comment is meant to imply that the code changes to make it happen are more complicated than seems reasonable.

PR #136, opened 2018-01-23
Migrate/hud and collections

Blue0ctober:
This removes the final files associated with the dashboard that lived under sensu-enterprise


asachs01:
LGTM 👍

PR #141, opened 2018-01-24
Fixes for Hugo >=0.32

### LINKED ISSUE #98: Upgrade Hugo
Hugo V.0.32.2 fails to generate content properly because they changed how layouts work a bit. I need to figure out what exactly changed and implement the changes so we can continue upgrading

dabria:
Closes #98 

Makes changes to files names, references different layouts. I've upgraded to the latest Hugo `(0.34)`

# Changes
- Any `index.md` must be renamed to `_index.md`
- index layouts in `layouts/_default` have been renamed to make sense
- Index pages require a layout!!! If you don't need a special layout in `layouts/default` then just use `layout: "single"`

# Requirements
- Everyone updates on merge
- Hugo version is updated in Heroku


cwjohnston:
in addition to updating Hugo in our heroku setup, we also need to update the Hugo executable in this repo which is used by Travis. The Travis build for this branch appears to have passed with the current version of Hugo.

dabria:
Some pages will fail to render on older versions, but there won't be an error. Any index with `layout: "single"` will not render properly on older versions.

cwjohnston:
Thanks @dabria, I have updated heroku to use 0.34.

PR #145, opened 2018-01-25
Re-add solaris postinst.sh

### LINKED ISSUE #142: Re-create Solaris 11 install script
We seem to have misplaced the Solaris 11 install script which creates a sensu user/group and chown/chmods a few directories. This script needs to be re-created so that users can install Sensu on Solaris 11.

amdprophet:
Fixes #142

### DIFF
--- a/static/files/postinst.sh
+++ b/static/files/postinst.sh
@@ -2,7 +2,7 @@
 
 # add sensu group
 set +e
-grep "^sensu:" /etc/group > /dev/null
+getent group sensu > /dev/null
 RC=$?
 set -e
 if [ "${RC}" != 0 ]; then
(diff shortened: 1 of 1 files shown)


static/files/postinst.sh
```diff
@@ -0,0 +1,28 @@
...
+grep "^sensu:" /etc/group > /dev/null
+RC=$?
+set -e
+if [ "${RC}" != 0 ]; then
+    groupadd sensu
+fi
+
+# add sensu group
+set +e
+grep "^sensu:" /etc/passwd > /dev/null
```
cwjohnston: We've previously been called out for grepping /etc/passwd instead of using `getent`. Is that applicable here as well?

cwjohnston:
Once this is merged, I think it should also be copied to the soon-to-be-former docs site.

PR #146, opened 2018-01-25
[WIP] Symlinks for latest

### LINKED ISSUE #86: Per-product redirects/aliases for "latest" docs
Under the existing middleman site we have a symlink for `latest` that points to the latest revision of the documentation. 

Ideally this site would have similar URLs, e.g. `/sensu-core/latest` or `/sensu-enterprise/latest`, which point to the most recent version of the docs as defined under `params.products.*.versions` in config.toml

From briefly reviewing hugo documentation it looks like aliases are often configured per-page using frontmatter, which doesn't seem practical for our case.

dabria:
Closes #86 

# Changes
Adds symlinks for projects we have content for.
- Sensu core
- Sensu enterprise
- Sensu enterprise dashboard
- Plugins

If your URL contains "latest" then the sidebar menu will be consistent and give you links that contain "latest", you can still navigate to the version numbers via the version dropdown.

# Issues
On server start, there are a significant amount of warnings about duplicate menu items. This is because Hugo sees the symlinked files (identified by title) as duplicates being added to the menu, rather than the same item. Hugo only recently allowed symlinks `0.32`, so I hope there's a way to suppress these eventually. Otherwise we would need to edit the front matter each time to change the menu they belong to (something like `product-latest`), which is not really sustainable. While these errors work locally, I don't believe they can be ignored on production.

screenshot 2018-01-25 11 05 48

I couldn't find any immediate solutions for this. Unsure if we should leave this for the time being, or just deal with the errors.

### DIFF
--- a/layouts/partials/nav.html
+++ b/layouts/partials/nav.html
@@ -67,7 +67,12 @@
             <!-- No children, this is top level menu items -->
             {{ else }}
               {{ if not (eq $.Params.product $x.Name) }}
-                <li><a class="menu-child" href="{{ .URL | absURL }}">{{ $x.Name }}</a>
+                {{ if in $.URL "latest" }}
+                  {{ $url := replace (.URL | absURL) $version "latest" }}
+                  <li><a class="menu-child" href="{{ $url }}">{{ $x.Name }}</a>
+                {{ else }}
+                  <li><a class="menu-child" href="{{ .URL | absURL }}">{{ $x.Name }}</a>
+                {{ end }}
               {{ end }}
             {{ end }}
             </li>


cwjohnston:
@dabria with the travis build failing on these changes, do we expect them to work when merged to master? Will the build be broken going forward?

dabria:
@cwjohnston I don't know for sure unfortunately. If my memory serves, I think there was a problem in the past where if there was an error it wouldn't even build on Heroku. That might just be related to our buildpack though? Sorry I'm not of much help here.

cwjohnston:
We've learned from the hugo community that symlinks won't work the way we want them to, in part because of how we have structured our menus.

This PR is superseded by #230 and https://github.com/sensu/heroku-buildpack-static/commit/ca2c0f210ad371ee5359ba92b6f6b1e202615599

PR #151, opened 2018-01-30
Fixes #150 where SSL config was missing. Also fixes broken link to reference doc on config merging.

### LINKED ISSUE #150: Add SSL Configuration Documentation for SED
Based on the uchiwa docs here: https://docs.uchiwa.io/guides/security/#tls-configuration we should be able to add this to the SED docs.

asachs01:


cwjohnston (approved):
LGTM

Issue #144, opened 2018-01-25
Adjust Documentation


mbbroberg:
Right before launching, we will need to: 

- [x] Adjust the language of the README after #143 
- [x] Move all the Hugo administration how-to documentation from the README.md to a better place (TBD)

Might be resolved with #90 but possibly separate.

cwjohnston:
Adding my two cents, we want to make sure this project's README refers to https://gohugo.io/getting-started/installing or similar for installation instructions, instead of duplicating that document here.

cwjohnston:
@mbbroberg any updates on this? #143 is merged and we removed Hugo installation comments in #157.

We may be able to completely remove comments about publishing to Heroku as that is [done automatically by Travis whenever we merge to master](https://github.com/sensu/sensu-docs-site/blob/master/.travis.yml#L9-L15).

cwjohnston:
I am going to move details related to Hugo formatting and layouts, Heroku configuration, etc into a Wiki on this repository.

mbbroberg:
Regarding the first checkbox: 

- We're still in need of a status update section until we're fully migrated to the new site (and old is properly deprecated) 
- The "Why Change" is still valid and could even use a word or two more to organize it

cwjohnston:
Marking the second check box as complete following #180

cwjohnston:
@mbbroberg is this closed by #231 ?

PR #180, opened 2018-02-12
Move chunks of Heroku and Hugo content to the Wiki

cwjohnston:
I've tried to move most of this content as-is to the wiki, and add more detail in the wiki as needed. Please have a look at both the deletions here and the wiki content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,11 +38,26 @@ cd sensu-docs-site ; git submodule update --init --recursive
 
 #### Installing Hugo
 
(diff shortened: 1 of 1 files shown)


cwjohnston:
Relates to #144

cwjohnston:
@mbbroberg added a section near the top with link to our wiki.

mbbroberg:
@cwjohnston perfect!

Issue #185, opened 2018-02-20
New changes to site are failing due to `yarn test` not being found


asachs01:
@grepory @jamesdphillips  I'm not exactly sure where to go on this one.  But based on https://travis-ci.org/sensu/sensu-docs-site/jobs/343926001 and #183  it seems like the index rebuilding is failing w/ Travis?

grepory:
I'll take a look today.

grepory:
Sorry about that @asachs01. #187 should take care of that.

PR #187, opened 2018-02-20
Add yarn test placeholder script

grepory:
Travis CI seems to run `yarn test` which I didn't expect. This adds a placeholder test script and the test script definition in `packages.json`.


grepory:
closes #185

PR #188, opened 2018-02-20
Update contributing

asachs01:
Can I get a 👍 @mbbroberg ?


mbbroberg (changes requested):
"written for all versions of the product you write it for" doesn't quite make sense to me. I think we want to ask people to submit a guide to a **specific** version and any other versions in which it applies. It'd be great to link to a commit that does this, though that doesn't have to gate us here. 

On the OS-specification, I thought CentOS was the default that's been discussed so far. If not, it'd do us well to take a step back and gauge what is best for users base by reviewing known environments and, if we don't have enough data, surveying users if needed.

mbbroberg (approved):
Looking good.

PR #193, opened 2018-02-22
Port hooks reference guide

### LINKED ISSUE #204: Reference/Hooks
Create a "Hooks" section in Sensu 2.0 "Reference": #149

nikkictl:
Signed-off-by: Nikki Attea <nikki@sensu.io>

Ports documentation on hooks.
Closes #204


content/sensu-core/2.0/reference/hooks.md
```diff
@@ -0,0 +1,96 @@
...
+menu:
+  sensu-core-2.0:
+    parent: reference
+---
+
+# Hooks
+
+### What are hooks?
+
+Hooks are commands run by the Sensu client in response to the result of a check, mutator, or handler execution. The Sensu client will execute the appropriate configured hook command, depending on the execution status (e.g. 1). The hook command output, status, executed timestamp, and duration are captured and published in the event result. Hook commands can optionally receive JSON serialized Sensu client data via STDIN.
```
palourde: > in response to the result of a check, mutator, or handler execution

I might be wrong but I believe hooks are only run in response to the result of a check command execution right now; I don't see any reference to hooks in the handler & mutator structs!

content/sensu-core/2.0/reference/hooks.md
```diff
@@ -0,0 +1,96 @@
...
+
+### New and improved hooks!
+
+In 2.x, we've redesigned and expanded on the concept of 1.x check hooks. Hooks are now their own resource, and can be created and managed independent of the check configuration scope. With unique and descriptive identifiers, hooks are now reusable and can respond to checks, mutators and handlers! And thats not all, you can now execute multiple hooks for any given response code.
```
palourde: > can respond to checks, mutators and handlers!

See previous comment

content/sensu-core/2.0/reference/hooks.md
```diff
@@ -0,0 +1,96 @@
...
+
+Check out Sean's [blog post](https://blog.sensuapp.org/using-check-hooks-a739a362961f) about Sensu Core check hooks to see how you can use Sensu for auto-remediation tasks!
```
palourde: 👍

palourde (changes requested):

content/sensu-core/2.0/reference/hooks.md
nikkiattea: Yes its not implemented yet! Do you think it would be reasonable to reference the feature tickets associated with mutator and handler hooks, or just remove that part completely?

echlebek: IMO we should remove references to things that are unimplemented, but create a TODO in the feature tickets that links back to this document for including them once they exist.

content/sensu-core/2.0/reference/hooks.md
```diff
@@ -0,0 +1,96 @@
...
+Organization:   default
+Environment:    default
+{{< /highlight >}}
+
+### Management
+
+Hooks can be configured both interactively or by using CLI flags.
+
+{{< highlight shell >}}
+sensuctl hook create nginx-start --command "sudo /etc/init.d/nginx start" --timeout 10
```
echlebek: _Very_ small quibble here, maybe we should encourage people to use single quotes around `--command` flags, since double quotes are subject to shell expansion. The case I'm thinking of here is a user copying an example from the docs and replacing the command with their own, inadvertently triggering shell expansion, and getting confused.

content/sensu-core/2.0/reference/hooks.md
```diff
@@ -0,0 +1,96 @@
...
+
+{{< highlight shell >}}
+sensuctl hook delete nginx-start
+{{< /highlight >}}
+
+## Check hooks
+
+### What is a check hook?
+
+A check hook is a type of hook that lives in the check configuration scope. Check hooks associate existing hooks to a check and type. The check hook type is defined by the response code or severity of the check execution result. Valid check hook types include (in order of precedence): “1”-“255”, “ok”, “warning”, “critical”, “unknown”, and “non-zero”.
```
echlebek: If "ok" is a valid hook type, then why isn't 0-255 allowed instead of 1-255?

echlebek (commented):
Very nice! Just a few questions and comments.

content/sensu-core/2.0/reference/hooks.md
nikkiattea: The reusability of hooks _allows_ us to add them to mutator and handlers. Otherwise, what's so different about 2.x hooks and why are we differentiating "hooks" and "check hooks." Just thinking out loud.. do you think the documentation is clear enough about that difference if we were to remove any references to mutator or handlers?

content/sensu-core/2.0/reference/hooks.md
nikkiattea: You bring up a good point! It's simply how 1.x is implemented, so I'm not sure why "ok" is the only way to represent a "0" response code.

echlebek: @grepory or @portertech can you comment on this?

grepory: @portertech can we revisit this interface?

portertech: The 1.x hook logic and config validator does allow 0-255, this is probably a doc oversight:

Hook code: https://github.com/sensu/sensu/blob/e895d5958f1a3857601d3ef321f59dfc4d9e8f1f/lib/sensu/client/process.rb#L133-L134

Validator code: https://github.com/sensu/sensu-settings/blob/aa818a30ae7ec9722929a3c5c6667f236b0ad05b/lib/sensu/settings/validators/check.rb#L201

nikkiattea: Oh funny. I’ll write up a ticket to make that quick change tomorrow!

nikkiattea: It was actually a doc oversight in 2.x too 🤣 Our regex includes 0, but I'm pushing a change to update our validation message with a minor bug fix!

```
// CheckHookRegexStr used to validate type of check hook
var CheckHookRegexStr = `([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])`
```

content/sensu-core/2.0/reference/hooks.md
palourde: We could definitively add a mention in the `New and improved hooks!` section mentioning that since hooks are now a distinct resource, it opens the door for their use with other type of resources than check in the future. Not mentioning _what_ resources and _when_ might help managing expectations!

palourde (approved):

PR #218, opened 2018-02-27
Grunt hugo-build command should fail if hugo fails

grepory:
We weren't checking the exit status of hugo when running it. I didn't even think about it. Also, grunt.util.spawn is async, so we have to make sure that we wait for the command to exit before returning from the task and proceeding to the next.

Also fatally exit if hugo fails.


Gruntfile.js
```diff
@@ -7,12 +7,20 @@ var CONTENT_PATH_PREFIX = "content/";
 module.exports = function(grunt) {
 
     grunt.registerTask("hugo-build", function() {
+        done = this.async();
```
jamesdphillips: Horrors of javascript's past still haunt it.. `done =` is valid but not good practice. It's not a big deal because of the scope of the script but I would change it to `const done = this.async();`

PR #232, opened 2018-03-06
Automatically install hugo when using yarn

cwjohnston:
This change set adds grunt tasks that use hugo-cli and grunt-env modules to achieve automatic installation of Hugo when running `yarn` commands.

As before, the version of Hugo to install is determined by the `HUGO_VERSION` environment variable. The default value of this variable is set to `0.34.0` in the Gruntfile.js, but can still be overridden at run time by setting a different value.

When run with default settings, this approach reduces potential for folks to be testing their changes against the wrong version of Hugo and removes the need for us to vendor the hugo binary in this repo.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -25,14 +25,14 @@ module.exports = function(grunt) {
         grunt.util.spawn({
             cmd: "hugo",
         },
-            function(error, result, code) {
-                if (code == 0) {
-                    grunt.log.ok("Successfully built site");
-                } else {
-                    grunt.fail.fatal(error);
-                }
-                done();
-            });
+        function(error, result, code) {
+            if (code == 0) {
+                grunt.log.ok("Successfully built site");
+            } else {
+                grunt.fail.fatal(error);
+            }
+            done();
+        });
     });
 
     grunt.registerTask("hugo-server", function() {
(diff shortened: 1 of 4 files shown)


Gruntfile.js
```diff
@@ -23,6 +35,60 @@ module.exports = function(grunt) {
...
+    });
+
+    grunt.registerTask("print-hugo-version", function() {
+        const done = this.async();
+        grunt.util.spawn({
+            cmd: "hugo",
+            args: ["version"],
+            opts: {stdio: 'inherit'}
+        },
+            function(error, result, code) {
```
echlebek: This appears to be tabbed over once too many times.

Gruntfile.js
```diff
@@ -23,6 +35,60 @@ module.exports = function(grunt) {
...
 
+    grunt.registerTask("hugo-server", function() {
+        const done = this.async();
+
+        grunt.log.writeln("Running Hugo server");
+        grunt.util.spawn({
+            cmd: "hugo",
+            args: ["server"],
+            opts: {stdio: 'inherit'}
+        },
```
jamesdphillips: Thought:

```
grunt.registerTask("hugo-server", function() {
  const done = this.async();
  const args = process.argv.slice(3); // fetch given arguments

  grunt.log.writeln("Running Hugo server");
  grunt.util.spawn({
    cmd: "hugo",
    args: ["server", ...args], // pass arguments down
    // ...
```

In this way we could pass any argument to hugo server obviating the need for the extra `hugo-server-fullrender` task and allowing the user to pass any additional arguments they might need.

```
yarn server --disableFastRender --veryVerboseOrSomething
```

Gruntfile.js
jamesdphillips: On a related note.. at some point we should add `prettifier` and an editor config to the project. 👍 Enforce consistent code automagically.

Gruntfile.js
cwjohnston: Thanks @jamesdphillips , I think this is a great idea. I'll give it a shot.

PR #233, opened 2018-03-07
Add filters docs for Sensu 2.0

palourde:
Closes https://github.com/sensu/sensu-docs-site/issues/201

The vast majority of this document has been copied from [1.x documentation](https://sensuapp.org/docs/1.2/reference/filters.html).

I'm also realizing that we will additional documentation on govaluate and improve the events reference documentation with an example of an event.


palourde:
This PR doesn't make sense anymore, we only need overview guides for now.

PR #237, opened 2018-03-09
Updates to why and how to get involved

### LINKED ISSUE #144: Adjust Documentation

mbbroberg:
As we get closer to launch, want to give more details on what we're up to!

Fixes #144


cwjohnston:
I'm not sure why this PR is in this state where we're only adding a single line. 

@mbbroberg will you please take a look at the current README in master and determine if the changes intended in this PR are already present in master?

mbbroberg:
@cwjohnston fixed it. This will need a fresh review.

README.md
```diff
@@ -45,6 +47,14 @@ cd sensu-docs-site ; git submodule update --init --recursive
 This project uses [Yarn](https://yarnpkg.com/) to manage dependencies and the build process. For installation
 on installing yarn, [view their documentation](https://yarnpkg.com/lang/en/docs/install/).
 
+#### Installing Hugo
```
cwjohnston: Installing `hugo` is now handled automatically when running `yarn` or `yarn server`, so we shouldn't need these instructions any longer.

mbbroberg:
@cwjohnston got it 👍

PR #248, opened 2018-03-14
Sensu go/silencing

### LINKED ISSUE #243: [Reference] Silencing

mercul3s:
Add a reference guide on silencing.

Closes #243

### DIFF
--- a/content/sensu-core/2.0/reference/silencing.md
+++ b/content/sensu-core/2.0/reference/silencing.md
@@ -122,20 +122,20 @@ example      | {{< highlight shell >}}
 
 organization | 
 -------------|------ 
-description  | the sensu RBAC organization that this check belongs to
+description  | the Sensu RBAC organization that this check belongs to
 required     | false 
 type         | String 
-default      | "default" 
+default      | current organization value configured for `sensuctl` (ie `default`) 
 example      | {{< highlight shell >}}
   "organization": "default",
 {{</ highlight >}}
 
 environment  | 
 -------------|------ 
-description  | the sensu RBAC environment that this check belongs to
+description  | the Sensu RBAC environment that this check belongs to
 required     | false 
 type         | String 
-default      | "default" 
+default      | current environment value configured for `sensuctl` (ie `default`) 
 example      | {{< highlight shell >}}
   "environment": "default",
 {{</ highlight >}}
--- a/content/sensu-core/2.0/reference/silencing.md
+++ b/content/sensu-core/2.0/reference/silencing.md
@@ -1,0 +1,126 @@
+---
+title: "Silencing"
+description: "The silencing reference guide."
+weight: 1
+version: "2.0"
+product: "Sensu Core"
+platformContent: true
+menu:
+  sensu-core-2.0:
+    parent: reference
+---
+
+# Silencing
+
+## How does silencing work?
+Silencing entries are created on an ad-hoc basis via `sensuctl`. When silencing entries are successfully created, they are assigned an `ID` in the format `$SUBSCRIPTION:$CHECK`, where `$SUBSCRIPTION` is the name of a Sensu agent subscription and `$CHECK` is the name of a Sensu check. Silencing entries can be used to silence checks on specific agents by by taking advantage of per-agent subscriptions, e.g. `agent:$AGENT_NAME`.
+
+These silencing entries are persisted in the Sensu data store. When the Sensu server processes subsequent check results, matching silencing entries are retrieved from the store. If one or more matching entries exist, the event is updated with a list of silenced entry ids. The presence of silencing entries indicates that the event is silenced.
+
+When creating a silencing entry, a combination of check and subscription can be specified, but only one or the other is strictly required.
+
+For example, when a silencing entry is created specifying only a check, its ID will contain an asterisk (or wildcard) in the `$SUBSCRIPTION` position. This indicates that any event with a matching check name will be marked as silenced, regardless of the originating agent’s subscriptions.
+
+Conversely, a silencing entry which specifies only a subscription will have an ID with an asterisk in the `$CHECK` position. This indicates that any event where the originating agent’s subscriptions match the subscription specified in the entry will be marked as silenced, regardless of the check name.
+
+## Silencing specification
+
+### Silenced entry ID 
+Silencing entries must contain either a subscription or check id, and are
+identified by the combination of `$SUBSCRIPTION:$CHECK`. If a check or
+subscription is not provided, it will be substituted with a wildcard (asterisk):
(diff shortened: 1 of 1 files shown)


content/sensu-core/2.0/reference/silencing.md
```diff
@@ -0,0 +1,196 @@
...
+platformContent: true
+menu:
+  sensu-core-2.0:
+    parent: reference
+---
+
+# Silencing
+
+## How does silencing work?
+Silencing entries are created on an ad-hoc basis via `sensuctl`. When silencing entries are successfully created, they are assigned an `ID` in the format `$SUBSCRIPTION:$CHECK`, where `$SUBSCRIPTION` is the name of a Sensu agent subscription and `$CHECK` is the name of a Sensu check. Silencing entries can be used to silence checks on specific agents by by taking advantage of per-agent subscriptions, e.g. `agent:$AGENT_NAME`.
```
palourde: Looks like we have a typo on that last line:

> agents by by taking

content/sensu-core/2.0/reference/silencing.md
```diff
@@ -0,0 +1,196 @@
...
+{
+  "expire": -1,
+  "expire_on_resolve": false,
+  "creator": null,
+  "reason": null,
+  "check": null,
+  "subscription": "agent:i-424242",
+  "id": "agent:i-424242:*"
+}
+{{ </highlight> }}
```
palourde: Looks like these spaces are causing a weird problem in my local installation of Hugo and the page won't stop loading. It probably need to be fixed for all examples.

screen shot 2018-03-14 at 10 08 14 am

content/sensu-core/2.0/reference/silencing.md
```diff
@@ -0,0 +1,196 @@
...
+  sensu-core-2.0:
+    parent: reference
+---
+
+# Silencing
+
+## How does silencing work?
+Silencing entries are created on an ad-hoc basis via `sensuctl`. When silencing entries are successfully created, they are assigned an `ID` in the format `$SUBSCRIPTION:$CHECK`, where `$SUBSCRIPTION` is the name of a Sensu agent subscription and `$CHECK` is the name of a Sensu check. Silencing entries can be used to silence checks on specific agents by by taking advantage of per-agent subscriptions, e.g. `agent:$AGENT_NAME`.
+
+These silencing entries are persisted in the Sensu data store. When the Sensu server processes subsequent check results, matching silencing entries are retrieved from the store. If one or more matching entries exist, the event is updated with a list of silenced entry ids. The presence of silencing entries indicates that the event is silenced.
```
palourde: We should probably indicate what happen when an event is silenced. For example, in 1.x docs, we have something like this:

> When the check name and/or subscription described in a silencing entry match an event, the handler will not be executed and an informational message will be logged.

content/sensu-core/2.0/reference/silencing.md
```diff
@@ -0,0 +1,196 @@
...
+### Silence all checks on entities with a specific subscription
+In this case, we'll compleltely silence any entities subscribed to `appserver`. Just as in the example of silencing all checks on a specific agent, we’ll create a silencing entry specifying only the `appserver` subscription:
+{{< highlight json >}}
+{
+  "subscription": "appserver", 
+}
+{{ </highlight> }}
+
+### Silence a specific check on entities with a specific subscription
+Assume a check `mysql_status` which we wish to silence, running on Sensu clients with the subscription `appserver`:
```
palourde: s/clients/entities/

content/sensu-core/2.0/reference/silencing.md
palourde: Additionally, I was thinking maybe we should replace `Sensu agent` with `Sensu entity` since it's the agent's entity that subscribes. What do you think?

content/sensu-core/2.0/reference/silencing.md
```diff
@@ -0,0 +1,196 @@
...
+regardless of subscriptions, we only need to provide the check name:
+{{< highlight json >}}
+{
+  "check": "mysql_status"
+}
+{{ </highlight> }}
+
+### Deleting silencing entries
+To delete a silencing entry, you will need to provide its id. Subscription only
+silencing entries:
```
palourde: That sentence sounds weird, but maybe it's just me haha:

> Subscription only silencing entries:

palourde (changes requested):
Awesome work, just some small details!

content/sensu-core/2.0/reference/silencing.md
mercul3s: Yes, let me add that, it should be the same behavior in 2.x. I'll double check that we do logging in that case.

content/sensu-core/2.0/reference/silencing.md
mercul3s: I agree, it makes more sense to refer to entities over agents. I'm already using that terminology other places in the doc, so I'll update it to be consistent.

mercul3s:
@palourde this is also ready for another round of review.

palourde (approved):

content/sensu-core/2.0/reference/silencing.md
```diff
@@ -0,0 +1,193 @@
...
+required     | false 
+type         | String 
+default      | null 
+example      | {{< highlight shell >}}
+  "creator": "mercedes",
+  "subsription": "webserver",
+  "check": "app_status",
+  "reason": "rebooting the world"
+{{</ highlight >}}
+
```
palourde: I'm just realizing now that we should probably also include the `environment` & `organization` attributes and I totally forgot to do that for filters!

mercul3s: Ah yeah, good point. I will add them here and to mutators as I don't think I did so either!

mercul3s: @palourde added org/env attributes. Can you double check the wording? Figure we can copy-pasta to other reference docs and keep it consistent :)

palourde: @mercul3s Maybe capitalize `Sensu`?

And about the default value, I believe it will use the current env/org from `sensuctl`, at least with the flags/interactive modes. I'm wondering if we should mark those as required in the docs 🤔

mercul3s: OK, I can make it more clear that it uses the current org/env from sensuctl. I think we document them as required and further describe that behavior when we get to the RBAC docs (and maybe also for setting up sensuctl), but not go into too much detail here. They are optional as long as the cli has been configured.

palourde: 👍 👍 👍

PR #254, opened 2018-03-15
Refactor the Sensu 2 installation docs

palourde:
- It adds the `Windows` platform for 2.0 docs.
- It slightly modifies the behaviour of `platformBlock` so multiple platforms can be defined and if one of them match the platform in the cookie, the `div` block is displayed. I.e., `{{< platformBlock "Ubuntu/Debian RHEL/CentOS" >}}` to match Linux, so we don't have to duplicate content between these two platforms.
- Reorganize the section with the changes mentioned above.
- Add a section for the services logs.

### DIFF
--- a/content/sensu-core/2.0/getting-started/installation-and-configuration.md
+++ b/content/sensu-core/2.0/getting-started/installation-and-configuration.md
@@ -139,8 +139,8 @@ Coming soon.
 Verify that the services are properly running using `journalctl`.
 
 {{< highlight shell >}}
-journalctl -u sensu-backend -f
-journalctl -u sensu-agent -f
+tail -f /var/log/sensu/sensu-backend.log
+tail -f /var/log/sensu-agent.log
 {{< /highlight >}}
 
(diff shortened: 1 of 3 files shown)


content/sensu-core/2.0/getting-started/installation-and-configuration.md
```diff
@@ -94,20 +115,51 @@ sudo /etc/init.d/sensu-backend start
...
+journalctl -u sensu-backend -f
+journalctl -u sensu-agent -f
+{{< /highlight >}}
+
+### Ubuntu 16.04 / CentOS 7 / RHEL 7
+
+Verify that the services are properly running using the log files.
+
+{{< highlight shell >}}
+tail -f /var/log/sensu/sensu-backend.log
```
mercul3s: We'll need to use journalctl for accessing logs in 16.04 as well.

palourde: Oh good catch, I'm realizing I have inverted these two sections!

PR #273, opened 2018-03-19
Mark active element in navigation

palourde:
This PR fix the navigation menu so the active page is properly identified in the menu and the parent menu is automatically open when the page is changed:

screen shot 2018-03-19 at 3 54 12 pm


cwjohnston:
Closes https://github.com/sensu/sensu-docs-site/issues/70

PR #280, opened 2018-03-20
Ubuntu systemctl enablement for RabbitMQ

mbbroberg:
Adding to not lose details from https://github.com/sensu/sensu-docs-legacy/pull/694

Let me know if this would be better added elsewhere in the docs.


asachs01 (approved):
👍 LGTM, also going to replicate this across other versions.

PR #290, opened 2018-03-22
Fix platforms filenames

jrRibeiro:
Fix the names of the Windows, FreeBSD and MacOS X installation files.

They are incorrect and causing 404s in the Downloads page - https://sensuapp.org/downloads - as well as in the Install Sensu Client pages (for all versions) - https://docs.sensu.io/sensu-core/1.2/installation/install-sensu-client/

Ex: https://docs.sensu.io/sensu-core/1.2/platforms/sensu-on-microsoft-windows/#sensu-core


cwjohnston:
I think this looks good, thank you @jrRibeiro ! 💚

cwjohnston (approved):

PR #302, opened 2018-03-26
remove duplicated sensu-doc-template.md, restore old path

cwjohnston:
In #258 we duplicated and moved the contents of `static/files` into each version of sensu-core documentation. In the case of `sensu-doc-template.md` this isn't really appropriate. This change set addresses that oversight and restores the file to it's original path so that CONTRIBUTING.md remains correct.


Blue0ctober (approved):
LGTM 👍

asachs01 (approved):
LGTM 👍

PR #303, opened 2018-03-26
[config] set enableRobotsTXT to true

cwjohnston:
We're currently logging a lot of 404s for `robots.txt`. This isn't functionally a problem but it would be nice to see fewer of these.

This change configures Hugo to generate its default `robots.txt` with the following content:

```
User-agent: *
```

This will allow any search engine to index the docs, functionally the same as having no `robots.txt`.

Per https://gohugo.io/templates/robots/ we can further customize this using a layout template, should the need arise.


Blue0ctober (approved):
LGTM 👍

Issue #196, opened 2018-02-22
Overview/Sending notifications with handlers


nikkiattea:
Create a "Sending notifications with handlers" section in Sensu 2.0 "Overview": #149

PR #306, opened 2018-03-28
Handlers guide

palourde:
Closes https://github.com/sensu/sensu-docs/issues/196


content/sensu-core/2.0/guides/send-slack-alerts.md
```diff
@@ -0,0 +1,105 @@
...
+### Installing the handler command
+
+The first step is to create an executable script named `handler-slack`, which is
+responsible for sending the event data to Slack. The source code of this script
+is available on [GitHub][3] and can easily be compiled using the [Go tools][4].
+The generated binary will be placed into one of the Sensu backend [`$PATH`
+directories][5], more precisely `/usr/local/bin`.
+
+{{< highlight shell >}}
+go build -o /usr/local/bin/handler-slack main.go
```
mercul3s: It would be good to have instructions for cross compiling this for different architectures, and maybe a pointer to installing Go as that would be required for a user to do this step. Oh also that they need to cd to handlers/slack/ to find the main.go file. These things will not be obvious to those unfamiliar with Go.

mercul3s (approved):
Test drove this and it worked for me! One minor note, otherwise looks good 🍕 🌮 🍨 🍔 🌯

content/sensu-core/2.0/guides/send-slack-alerts.md
palourde: Done!

Issue #297, opened 2018-03-24
Both dropdown options are initially shown for platform


majormoses:
On [this page](https://docs.sensu.io/sensu-core/1.2/reference/ssl/#generate-self-signed-openssl-certificates-and-ca) you have a dropdown to switch the platform but it initially shows up n times where n is the number of platforms.

![image](https://user-images.githubusercontent.com/3145127/37866471-59c65bb4-2f48-11e8-8d05-5cef67b3db7b.png)

After selecting a platform the others go away and default to the platform you have previously suggested (had to go incognito to replicate to avoid cache). If its not a bug it's certainly an odd quirk.

![image](https://user-images.githubusercontent.com/3145127/37866497-9c317632-2f48-11e8-8d97-7f6f03fe4142.png)

majormoses:
While its clearly not an issue with a small number of platforms it could get very annoying if we had a half dozen or more.

hillaryfraley:
Team has fixed this

majormoses:
am still able to replicate this it initially shows both versions as per the screenshot, after selecting one it works fine. I would suggest testing this in an incognito window as I was able to reproduce but then hitting the page again I was not able to reproduce without an incognito window.

hillaryfraley:
@majormoses I believe this is now fixed via https://github.com/sensu/sensu-docs/pull/3047, but can you confirm whether you are still seeing the problem?

PR #307, opened 2018-03-28
Allow platforms override per page

palourde:
This PR introduces the concept of platform override per page in front matter, which allows new platforms to be introduced for certain pages. By doing so, it prevents the introduction of a platform (e.g., Windows) across all documentation if we only need it for few particular pages.

If the platform cookie is set to a particular platform and the user browse to a subsequent page where this platform does not exit, the cookie gets removed so we get back to the `All Platforms` option and all content is displayed.

I also did some cleanup.


static/js/platformDropdowns.js
```diff
@@ -39,19 +42,14 @@ function setHeaderDropdowns($uniqueness) {
...
+function setPlatformCookie($platform) {
   Cookies.set('platform', $platform, { expires: $cookieLife });
   togglePlatformDivs();
   setTopDropdown();
-  $platform = $platform.replace("/", "_");
-  $platform = $platform.replace(" ", "-");
-  console.log($platform)
-  var dropdown = document.getElementById("platformButtonTitleText-" + $platform);
-  dropdown.innerHTML = Cookies.get("platform") + '<i class="fa fa-chevron-down dropdownArrow-' + $platform + '" aria-hidden="true"></i>';
-  dropdown.scrollIntoView();
```
palourde, line 55: These 6 lines were causing an error in the browser console because `dropdown` was `null`, so I simply removed it.

layouts/partials/platformDropdown.html
```diff
@@ -11,27 +11,28 @@
...
+          <a class="platformButtonTitle"><span id="platformButtonTitleText">All Platforms</span><i class="fa fa-chevron-down dropdownArrow" aria-hidden="true"></i></a>
           <div class="hiddenPlatforms" style="display: none;">
  
             <!-- List of platforms -->
             <ul class="hiddenItems">
-              {{ if isset $product_info.versions .Params.version }}
-                {{ range (index $product_info.versions .Params.version) }}
-                  {{ $platform := . }}
-                  {{ $title2 := replace $.Title " " "-" }}
-                  {{ $title2 := lower $title2 }}
```
palourde, line 32: Seems like `$title2` was never used.

dabria (approved):

cwjohnston (commented):
I haven't had a chance to review this in a browser but I am wondering if this approach allow us to remove platforms from config.toml entirely? Seems like a good idea...

palourde:
@cwjohnston We could effectively remove the platforms in `config.toml` but if we do so, we should probably cleanup the code and we will need to explicitly define those on pages that use platform blocks (I don't know if they are widely used in 1.x docs though).

asachs01:
I'll start reviewing the 1.x docs to see where we're actively using the platforms feature.

palourde:
FYI, we could first merge this PR and then progressively move the platforms to each page that require them so we don't have to modify everything at once.

asachs01:
Adding #297 here for reference

cwjohnston:
@palourde sorry I was blocking this for so long, my mistake! 🙈

PR #318, opened 2018-04-04
Securing Sensu Guide

asachs01:
_**Finally**_ getting this guide added.


majormoses (commented):
First off thank you for writing this up, it really has been a long time coming. Here are some additional things to discuss when talking about securing sensu:
- using `safe_mode`: sensu is basically a remote command and control platform. By not locking this down you give attackers the perfect platform to attack as this allows them to easily spread through the network and how this can be easily mitigated. 
- using sudo checks without giving away too much access
- disabling http and tcp sockets
- while you briefly mention redacting it might be good to spend a little bit more time explaining why you should use token substitution and redaction together. Also I know it has been a gotcha that when updating the array of redacted keys that defining it does not append but replaces the original array so you should always include the default keys unless you specifically want to leak a particular secret. Not sure why you would want to do that other than possibly a demo on sensu security.

I am not saying you need to discuss these but these are important topics to discuss.

asachs01:
Awesome @majormoses ! Thanks so much for the feedback! I'll add those to the guide.

content/sensu-core/0.29/guides/securing-sensu.md
```diff
@@ -0,0 +1,303 @@
...
+* How to secure Uchiwa/the Sensu Enterprise Dashboard
+
+We'll also walk through securing the additional components like RabbitMQ and Redis in the guides following this one.
+
+# Objectives
+
+We'll cover the following in this guide:
+
+* [Securing Sensu](#securing-sensu-clients)
+  * [The `client_signature` Attribute](#the-clientsignature-attribute)
```
cwjohnston: Ensuring authenticity of check results using client signatures

content/sensu-core/0.29/guides/securing-sensu.md
```diff
@@ -0,0 +1,303 @@
...
+  * [The `redact` Attribute](#the-redact-attribute)
```
cwjohnston: Preventing inadvertent disclosure of secrets using client-side redaction

content/sensu-core/0.29/guides/securing-sensu.md
```diff
@@ -0,0 +1,303 @@
...
+  * [SSL/TLS Encryption for Transport Communication](#client-ssl-tls)
```
cwjohnston: Encrypting communications using SSL/TLS

content/sensu-core/0.29/guides/securing-sensu.md
```diff
@@ -0,0 +1,303 @@
...
+* [Securing Sensu Server/API](#securing-sensu-server)
```
cwjohnston: IMO the SSL/TLS recommendations for transport (AMQPS for sensu-client, sensu-server, sensu-api and/or sensu-enterprise) and web UI (HTTPS for sensu-api+uchiwa or sensu-enterprise+sensu-enterprise-dashboard) should all go under the same "Encrypting communications using SSL/TLS"

cwjohnston (changes requested):
I think this is a good start, but I believe that we need to reorganize the headings a bit so that its clearer to users how this information helps them address a particular aspect of security (problem space) as opposed headings which describe security-related features of Sensu (solution space).

majormoses:
@asachs01 came back here and realized I mixed up something I originally conveyed which I have fixed. I think there should be a section talking about `safe_mode` and how it can prevent someone from gaining access to sensu and then using it to spread to other systems by executing arbitrary code as checks. Safe mode really adds a ton of protection and is often looked over. A non security note on them is that it can also help reduce false positives while nodes are bootstrapping. The check dependencies may not be installed before the client is installed.

content/sensu-core/0.29/guides/securing-sensu.md
asachs01: ✔️

content/sensu-core/0.29/guides/securing-sensu.md
asachs01: ✔️

content/sensu-core/0.29/guides/securing-sensu.md
asachs01: ✔️

content/sensu-core/0.29/guides/securing-sensu.md
asachs01: ✔️

asachs01:
@majormoses I'll go ahead and add that to the article:

- [x] using safe_mode: sensu is basically a remote command and control platform. By not locking this down you give attackers the perfect platform to attack as this allows them to easily spread through the network and how this can be easily mitigated.
- [x] disabling http and tcp sockets
- [x] while you briefly mention redacting it might be good to spend a little bit more time explaining why you should use token substitution and redaction together. Also I know it has been a gotcha that when updating the array of redacted keys that defining it does not append but replaces the original array so you should always include the default keys unless you specifically want to leak a particular secret. Not sure why you would want to do that other than possibly a demo on sensu security.

After some discussion w/ @cwjohnston , we're going to do a separate guide on sudo checks. 

I think I've included all the feedback you've mentioned, if you want to 👁 at it.

content/sensu-core/0.29/guides/securing-sensu.md
```diff
@@ -0,0 +1,319 @@
...
+  * [Securing Uchiwa](#securing-uchiwa)
+  * [Securing Sensu Enterprise Dashboards](#securing-sensu-enterprise-dashboards)
+
+## Securing Sensu Clients
+
+### Ensure Check Result Authenticity Using Client Signatures{#the-clientsignature-attribute}
+
+By default, the `signature` attribute isn't required and doesn't have a value, but adding it to your clients' configurations ensures that you're able to validate the results coming from a client by providing a random string in the attribute. In this way, if you're ever in doubt about the authenicity of the results that are being returned from a client, you can compare the results against your client configuration to see if the signature strings match up. You can read more about using a client `signature` [here][1]
+
+### Prevent Secret Discosure via Client-side Redaction{#the-redact-attribute}
```
palourde: s/Discosure/Disclosure/

cwjohnston (approved):

PR #324, opened 2018-04-09
[WIP] Scaling Sensu Guide

asachs01:
Could use some 👁 and feedback on this. It feels less like a guide and more of a discussion


asachs01:
@apaskulin I wonder if you might have some suggestions on this article. I'm not in love with how the legacy doc this was pulled from was written. My feeling is that with some diagrams, this could be particularly useful, but the initial example from failshell is definitely a legacy artifact.

apaskulin:
There's some good content here, but I agree that it could be more user friendly. I think diagrams (and ideally, case studies) will be the most helpful in illustrating the different options. I'm always happy to digitize diagrams if you sketch them out.

- Link to failshell article doesn't seem to work. Found the content [here](https://gist.github.com/warmfusion/f8e5a85ed28f63190a9b).
- Does this apply only to users with their own physical servers? Assuming there are more ways to scale Sensu than the options described in this guide, can we add some general language around considerations and constraints that may apply across different options?
- For little things like formatting and style changes, would you prefer comments or changes to the branch?

asachs01:
@sensu/success I could use some eyes on this again. I'm feeling a bit stuck with the article, and I'll see if I can give some adequate voice to where I feel stuck. Here's a list of things in my head as I'm reading/writing this doc.

* Is this a guide, or something that should live under `Installation` with the installation strategy doc?
* We already discuss setting up HA configurations for the various components (save the API), but they're in various places in the docs. Should this focus more on consolidating those resources?
* Could this be 2 different docs (scaling strategies vs. scaling individual components)

There's a thought that we could just cut out the individual component stuff I've added, and instead stick to discussing the strategies, and not do a "guide" style doc at all. I'm just not sure of the direction and aim of the doc, and it feels kind of like the "securing Sensu" guide did originally in that there was too broad a scope, but I'd like to get someone else's take.

rgeniesse:
- Might actually be a good spot for the installation area, so people know that HA us a possibility from the get go. Either that or a link to the guide section.
- I'd probably have to see all those locations. However, this may be a larger issue with any doc site as it grows and might be part of @apaskulin effort on documentation content epic.
- Maybe, but my comment below sorta hits on this.

I think scaling individual components itself is a strategy and showcases the agility(?) of Sensu 1.x and its components. Scope creep is real and maybe we would all benefit from agreeing on what guides in general should cover before someone starts to work on them. I've felt it on the monitoring the monitor guide.

asachs01:
Based on the convo today, I'm going to add subsequent guides that split this all out a bit better

PR #325, opened 2018-04-09
Artical add article pagination

inghamemerson:
Re-adding article pagination that was erroneously removed in a previous PR.


cwjohnston (approved):
LGTM! Thanks for the quick turn around.

PR #328, opened 2018-04-10
Sensu go/entities

### LINKED ISSUE #265: [Reference] Entities
The entities reference documentation needs a refactoring!

mercul3s:
Closes #265


content/sensu-core/2.0/reference/entities.md
```diff
@@ -4,82 +4,355 @@ description: "The entities reference guide."
...
+## New and improved entities
 
-## Agent entities
+An `entity`, formally known as a `client` in Sensu 1.x, represents anything (e.g. server, container, network switch, etc) that needs to be monitored. Sensu 2.0 uses an updated data model that allows for it to acknowledge the full range of infrastructure, runtime and application types that compose a complete monitoring environment (from server hardware to "serverless" functions). Sensu no longer focuses on the object doing the monitoring and instead focuses on the object it monitors. We call these monitored parts of an infrastructure "entities." An Entity not only provides context to event data (what/where the event is from) but an event's uniqueness is determined by the check name and the ID of the entity upon which the check ran. In addition, an entity can contain system information such as the hostname, OS, platform, and version.
```
palourde: s/Entity/entity/

content/sensu-core/2.0/reference/entities.md
```diff
@@ -4,82 +4,355 @@ description: "The entities reference guide."
...
+type         | string 
+example      | {{< highlight shell >}}"ID": "example-hostname"{{</ highlight >}}
 
-To create an agent entity, `sensu-agent` must be installed. Upon spinning up the sensu-agent, you can use several flags to customize the properties of the entity. In the example below, the optional flags are used to describe the default values of the agent entity that gets created upon starting sensu-agent. Other flags/properties you can set for your agent entity include: `--deregister`, `--subscriptions`, `--deregistration-handler`, and `--redact`.
+class        | 
+-------------|------ 
+description  | The entity type, validated with go regex [`\A[\w\.\-]+\z`](https://regex101.com/r/zo9mQU/2)
+required     | true
+type         | string 
+example      | {{< highlight shell >}}"class": "agent"{{</ highlight >}}
```
palourde: We should probably list the allowed values, `agent` & `proxy`!

content/sensu-core/2.0/reference/entities.md
```diff
@@ -4,82 +4,355 @@ description: "The entities reference guide."
...
 
-{{< highlight shell >}}
-sensu-agent start --id hostname --environment default --organization default --keepalive-timeout 120 --user agent
-{{< /highlight >}}
+subscriptions| 
+-------------|------ 
+description  | A list of subscription names for the entity. 
+required     | false 
+type         | array 
+default      | The entity-specific subscription.
```
palourde: Could we give an example or the format specs, i.e., `entity:{ID}`?

content/sensu-core/2.0/reference/entities.md
```diff
@@ -4,82 +4,355 @@ description: "The entities reference guide."
...
+example      | {{< highlight shell >}}"keepalive_timeout": 120 {{</ highlight >}}
 
-To view all the entities that are currently configured for the cluster, enter:
+organization | 
+-------------|------ 
+description  | The Sensu RBAC organization that this entity belongs to.
+required     | false 
+type         | string 
+example      | {{< highlight shell >}}
+  "organization": "default"
```
palourde: I just realized the spaces at the beginning of this line change the indentation of the examples for the org & env:

screen shot 2018-04-11 at 10 20 49 am

content/sensu-core/2.0/reference/entities.md
```diff
@@ -4,82 +4,355 @@ description: "The entities reference guide."
...
+-------------|------ 
+description  | Custom attributes to include as with the entity, that appear as outer-level attributes.
+required     | false 
+type         | byte 
+example      | {{< highlight shell >}}"team":"ops"{{</ highlight >}}
 
-### Management
+redact       | 
+-------------|------ 
+description  | List of items to redact from log messages.
```
palourde: We should probably specify that when using this attribute, the default values get overridden and not appended.

content/sensu-core/2.0/reference/entities.md
```diff
@@ -4,82 +4,355 @@ description: "The entities reference guide."
...
+        }
+      ]
+}
+{{</ highlight >}}
 
-{{< highlight shell >}}
-sensuctl entity delete entity.localhost
+arch         | 
+-------------|------ 
+description  | The entity's system architecture.
```
palourde: We should probably specify that we use the binary architecture (`runtime.GOARCH`) to determine the system architecture, meaning if an `amd64` system runs the `386` binary, it will be reported as `386` in Sensu!

For reference: https://github.com/sensu/sensu-go/issues/154

palourde (changes requested):
Thank you for this tremendous work 🙏🙏 🙏 

Just some minor concerns, otherwise it's perfect!

content/sensu-core/2.0/reference/entities.md
grepory: So entity type is actually sort of confusing.

An agent entity is a reserved word and should only be used by sensu agents.

I can’t actually find anywhere in the code that a proxy entity type means anything via  simple search, and I can’t think of anything off the top of my head.

People can put whatever value they want here—but if they put agent, then it invokes very specific functionality on the backend.

Honestly. We should rethink this field. The agent entity type is a very opaque feature.

content/sensu-core/2.0/reference/entities.md
```diff
@@ -4,82 +4,355 @@ description: "The entities reference guide."
...
 
-### What is a Sensu entity?
+Proxy entities (formerly known as proxy clients, "Just-in-time" or "JIT" clients) are dynamically created entities, added to the entity store if an entity does not already exist for a check result. Proxy entity registration differs from keepalive-based registration because the registration event happens while processing a check result (not a keepalive message). Sensu proxy entities allow Sensu to monitor external resources (e.g. on systems and/or devices where a sensu-agent cannot be installed, such a network switches), using the defined check ProxyEntityID to create a proxy entity for the external resource. Once created, proxy entities work much in the same way as any other Sensu entity.
 
-An **entity**, formally known as a client in Sensu 1.x, represents anything (e.g. server, container, network switch, etc) that needs to be monitored. It can contain system information such as the hostname, OS, platform, and version. An Entity not only provides context to event data (what/where the event is from) but an event's uniqueness is determined by the check name and the ID of the entity upon which the check ran.
+## New and improved entities
```
grepory, line 21: It seems like this should come before proxy entities?

content/sensu-core/2.0/reference/entities.md
grepory: I don’t think there are actually allowed values here, but agent and proxy are reserved. Agent for sensu-agent processes and proxy for proxy entities.

palourde: You are totally right, it's actually not enforced, my bad!

mercul3s: Should I add `agent` as a default value?

content/sensu-core/2.0/reference/entities.md
mercul3s: absolutely.

content/sensu-core/2.0/reference/entities.md
mercul3s: 🤦‍♀️  I don't even know why those had linebreaks and spaces. Dang you copy pasta!

content/sensu-core/2.0/reference/entities.md
mercul3s: Oh, that is something I wasn't aware of - will add a note.

content/sensu-core/2.0/reference/entities.md
mercul3s: I'm fine with moving it, but was following the convention for where "New and improved ..." falls in the current 2.0 documentation (it's after the "how x works" sections). If we move it here we might want to move it for everything else too.

content/sensu-core/2.0/reference/entities.md
mercul3s: 👍

content/sensu-core/2.0/reference/entities.md
palourde: Yeah, I'm still not sure where this section should placed 🤔

content/sensu-core/2.0/reference/entities.md
palourde: AFAIK the agent will set that value for its own entity and use `proxy` for result submitted through the socket & proxy checks, but otherwise it's not enforced by the API/backend. As Greg mentioned in its deleted comment, maybe the role of this attribute is not clear enough?

mercul3s: Ok that makes sense. I'll clarify that this value is set by the agent/proxy.

content/sensu-core/2.0/reference/entities.md
mercul3s: I think we should leave these sections later in the document to avoid confusing users new to Sensu with potentially two sets of vocabulary.

mercul3s:
@palourde feedback addressed!

palourde (approved):

PR #286, opened 2018-03-21
Updating Docs to ES Styles

inghamemerson:
This includes a number of updates to the docs for styles and some layout updates. See some screenshots below.

Thank you @mmosholder for her work on this!

screen shot 2018-03-21 at 1 23 48 pm
screen shot 2018-03-21 at 1 23 57 pm
screen shot 2018-03-21 at 1 24 15 pm
screen shot 2018-03-21 at 1 27 29 pm
screen shot 2018-03-21 at 1 27 21 pm

PR #261, opened 2018-03-18
add github live editor link

cwjohnston:
Adds a link to the left side drawer, inviting users to submit pull requests via GitHub's live editor.

PR #334, opened 2018-04-11
restore "Edit this page" link in drawer partial

cwjohnston:
Originally added in #261,  this change restores the "Edit this page" functionality we lost by merging #286.


asachs01 (approved):
👍 LGTM

PR #335, opened 2018-04-11
Return 302 when redirecting to /plugins/latest

cwjohnston:


obfuscurity:
We should be doing 302's for all of the `/latest` rewrites imho.

cwjohnston:
Closing in favor of https://github.com/sensu/sensu-docs/pull/345

PR #348, opened 2018-04-13
Dismissible & customizable alert banner

### LINKED ISSUE #94: Dropdown or section on top site
There needs to be space for a narrative so when people go to `docs.sensu.io` we can provide messaging about the state of migration. That can come in the form of either a text section or a dropdown similar to our current site (the "Join Slack" one). Functional requirements should include a single file edit to:

* Turn the message on or off
* Customize the message
* Customize the color of the background and text (particularly green/white for positive notes, yellow/charcoal for warnings and soft red/white for warnings)

Unless the idiomatic way to do so in Hugo is in multiple places. 🤷‍♂️ 

Example: 

screen shot 2017-12-20 at 2 27 58 pm

dabria:
## Details
Closes #94 (finally)

Adds:
- Ability to dismiss alert
- Ability to pass text, colour, and optionally a link to the partial

We now have customizable banners! (In both partial and shortcode form)
```
{{ $message := "WOW! That's neat!" }}
{{ $link := "Any url you like! You don't even NEED one!" }}
{{ $params := slice "#cc7d7c" $message $link }}
{{ partial "alert" $params }}
```

Just pass colour, message, and OPTIONALLY a link (in that order)!
```
{{ $color := index . 0 }}
{{ $message := index . 1 }}
{{ $link := index . 2 }}
```
They stack and unstack! Wow! That was actually kinda tricky to program them to stack properly!

Single alert with link
image

Multiple alerts
image

Dismissed alert
image

### DIFF
--- a/layouts/partials/alert.html
+++ b/layouts/partials/alert.html
@@ -5,8 +5,17 @@
 <div class="alertBanner" style="background-color: {{ $color }}">
   {{ if $link }}
     <a class="alertBannerText" href="{{ $link }}">{{ $message }}</a>
+    <span class="closeAlert"><i class="fa fa-times" aria-hidden="true" onClick="closeAlert()"></i></span>
   {{ else }}
     <span class="alertBannerText">{{ $message }}</span>
+    <span class="closeAlert"><i class="fa fa-times" aria-hidden="true" onClick="closeAlert()"></i></span>
   {{ end }}
 </div>
 
+
+<script>
+  function closeAlert() {
+    var $banner = document.getElementsByClassName("alertBanner");
+    $banner[0].style.display = 'none';
+  }
+</script>
--- a/layouts/partials/alert.html
+++ b/layouts/partials/alert.html
@@ -2,11 +2,11 @@
 {{ $message := index . 1 }}
 {{ $link := index . 2 }}
 
-<div class="versionBanner" style="background-color: {{ $color }}">
+<div class="alertBanner" style="background-color: {{ $color }}">
   {{ if $link }}
-    <a class="versionBannerText" href="{{ $link }}">{{ $message }}</a>
+    <a class="alertBannerText" href="{{ $link }}">{{ $message }}</a>
   {{ else }}
-    <span class="versionBannerText">{{ $message }}</span>
+    <span class="alertBannerText">{{ $message }}</span>
   {{ end }}
 </div>
 
--- a/layouts/partials/alert.html
+++ b/layouts/partials/alert.html
@@ -18,11 +18,13 @@
 <script>
   alerts = Array.prototype.slice.call(document.getElementsByClassName("alertBanner"));
   currentBanner = $(document.getElementById({{ $uniqueness }}));
+  // 10px is a small hack for different screen sizes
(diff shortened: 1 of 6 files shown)


layouts/shortcodes/alert.html
```diff
@@ -0,0 +1,49 @@
+{{ $color := index . 0 }}
```
palourde: Any reason why we need to duplicate the `alert.html` across the `shortcodes` and `partials` folder? I might be wrong but I have a feeling this particular one is not required.

palourde:
FYI, currently in review!

layouts/partials/alert.html
```diff
@@ -0,0 +1,49 @@
...
+  currentBanner = $(document.getElementById({{ $uniqueness }}));
+  // 10px is a small hack for different screen sizes
+  // semi-breaks if you're scaling the page large to small, but is fine on refresh
+  height = currentBanner.height() + 10;
+
+  // stack the alerts as needed
+  for (i = 1; i <= alerts.length; i++) {
+    lastLocation = parseInt($(alerts[i]).css('top'), 10);
+    newLocation = lastLocation + height;
+    currentBanner.css('top', newLocation + "px");
```
palourde: Maybe we should a parent section named `alertBanners`, which contains the `width`, `top` & `position` properties and we just add `alertBanner` div elements to that section so we don't have to deal with this calculation.

static/stylesheets/override.css
```diff
@@ -916,7 +931,7 @@ button.accordion.active, button.accordion:hover {
   //  margin-left: 0;
   //}
 
-  //.versionBanner {
+  //.alertBanner {
```
palourde: Maybe we could simply remove that code since it's commented?

layouts/partials/alert.html
palourde: Additionally, maybe it would be easier to add the proper padding to that parent section so it does't go over the page navigation/title, like in this screenshot:

palourde (changes requested):
Awesome work 🎉

Just some small concerns, otherwise it's perfect!

layouts/shortcodes/alert.html
dabria: My main reasoning was that we might want to call a banner on a specific page inside the page's markdown file.

I have looked into it, but from my understanding you can't call a partial from markdown, and you can't call a shortcode from a layout.

Rather than duplicating the code though, I'll instead have the shortcode call the partial. Keeping both accessible, but removing that duplication (which I hated anyway)

layouts/partials/alert.html
dabria: Great idea!

dabria: You pointed me on such a good path here! It felt SO good to remove all that janky math.

palourde:
Looks like it appears on the home page and I assume it's not desired!
screen shot 2018-04-17 at 9 26 30 am

palourde (approved):
Maybe just fix the homepage, otherwise it's perfect, awesome work!

dabria:
Good catch! I've added a check.

Issue #178, opened 2018-02-07
Explore merging Uchiwa content into Docs Site


asachs01:
Since Uchiwa uses Hugo and is already written in Markdown, it seems like a fairly simple thing for us to bring the content into the Docs site, and maybe do a redirect from the Uchiwa site to our docs.

mbbroberg:
This aligns well with consolidation across other microsites (sensu-plugins.io).

mbbroberg:
Note that Simon is open to it - we spoke about it today.

asachs01:
#349 is the PR for this

cwjohnston:
Closed by #349

PR #349, opened 2018-04-13
Uchiwa/port over existing uchiwa docs

asachs01:
Addresses #178


apaskulin:
This content looks good to me! Just not sure where we would set up the redirects. I wasn't able to find the source for https://uchiwa.io/#/

asachs01:
@cwjohnston have we decided if we're going to pull things in?

asachs01:
Going to merge this per @cwjohnston and then get the rest of the info figured out as far as changing the dns stuff w/ @palourde

cwjohnston (approved):

PR #363, opened 2018-04-16
Updating frontmatter in indexes

asachs01:
Frontmatter had incorrect "parent" attribute listed in indexes. Updated to correct parent.


Blue0ctober (approved):
LGTM 👍

Issue #337, opened 2018-04-11
[Reference] Time Windows


palourde:
Used by checks `subdue` attribute and filters `when` attribute.

PR #366, opened 2018-04-17
Add check subdues guide

palourde:
Closes https://github.com/sensu/sensu-docs/issues/337.


mercul3s (changes requested):
This worked for me, but I would suggest adding a note about timezones here. My test server runs in UTC, and I'd need to offset the time to equate for my business hours UTC offset (-7h).

palourde:
@mercul3s Oh yeah, totally forgot about those, good catch!

palourde:
@mercul3s Example updated!

mercul3s (approved):

PR #372, opened 2018-04-24
Add hardware recommendation guide.

echlebek:
This commit adds some documentation about minimum and recommended
hardware configurations.

Signed-off-by: Eric Chlebek <eric@sensu.io>

## Motivation and Context
To effective deploy Sensu Core 2.0, users will need to meet the minimum hardware requirements.

## How Has This Been Tested?
Extensive load testing has been performed to determine the performance characteristics of the Sensu backend and agent.

## Types of changes
- [x] New doc/guide


content/sensu-core/2.0/getting-started/recommended-hardware.md
```diff
@@ -0,0 +1,78 @@
+---
+title: "Hardware Requirements and Recommendations"
+linkTitle: "Hardware Requirements"
+description: "Guidelines for what you'll need for deploying Sensu Core to production."
+weight: 1
+version: "2.0"
+product: "Sensu Core"
+platformContent: true
+platforms: ["Ubuntu/Debian", "RHEL/CentOS", "Windows"]
```
palourde: We don't actually use the platform content on this page so `platformContent` should probably be set to `false` and we can remove this line!

asachs01:
Ohhh! Nice! Is this relevant for 1.x as well?

asachs01:
Also, I'm curious about the testing methods--could one of us on @sensu/success do testing on 1.x if this isn't relevant across product versions?

content/sensu-core/2.0/getting-started/recommended-hardware.md
```diff
@@ -0,0 +1,78 @@
...
+Each check that is executed will result in writes occurring in the backend.
+When provisioning storage hardware, a good rule of thumb is to have twice as
+many iops as you expect to have checks executing per second. Don't forget to
+include keepalives in this calculation; each agent will execute a keepalive
+every 20 seconds. So in a cluster of 100 agents, you can expect those agents
+to consume 10 write iops for keepalives.
+
+The Sensu Backend uses a relatively modest amount of RAM under most
+circumstances, but some users with a large amount of information stored in etcd
+may find that Sensu will mmap a large amount of RAM (2-4 GB). In almost all
```
palourde: Maybe replace `mmap` with something in plain english for non-technical users!

content/sensu-core/2.0/getting-started/recommended-hardware.md
```diff
@@ -0,0 +1,78 @@
...
+platforms: ["Ubuntu/Debian", "RHEL/CentOS", "Windows"]
+menu:
+  sensu-core-2.0:
+    parent: getting-started
+---
+
+### Backend Minimum Requirements
+
+* 64-bit Intel or AMD CPU
+* 4 GB RAM
```
palourde, line 17: Out of curiosity, where this number comes from? A first glance, it seems a lot but I'm probably missing something.

palourde (dismissed):
Great work!! Just some small concerns.

cwjohnston:
@asachs01 my understanding is that Sensu 1.x event processing isn't parallelized across multiple CPU cores the way Sensu 2.x is, so while the backend minimum and recommended specs provided here are probably suitable starting points for 1.x, I don't think we can yet make recommendations as specific as these, e.g. scaling the number of CPU cores to match an increase in the checks-per-second.

If the engineering team can help us find a reasonable way to reproduce the 2.x load testing methodology for 1.x, that would be very helpful to create similar recommendations for 1.x customers.

content/sensu-core/2.0/getting-started/recommended-hardware.md
echlebek: It's a compromise between my load testing observations and this passage from the etcd docs:

>  etcd has a relatively small memory footprint but its performance still depends on having enough memory. An etcd server will aggressively cache key-value data and spends most of the rest of its memory tracking watchers. Typically 8GB is enough. For heavy deployments with thousands of watchers and millions of keys, allocate 16GB to 64GB memory accordingly.

echlebek:
@asachs01 very much not relevant for 1.x. All of these recommendations were derived from load testing 2.x and the etcd documentation. I haven't done any load testing of 1.x so I can't give you any estimates. (I wonder if someone in the community may have?)

As far as load testing methodology goes, we would also need to come up with something new for 1.x, as 1.x depends on Redis and RabbitMQ and 2.x does not.

echlebek:
I'd like to expand a bit on my last comment.

Because we have these other external service dependencies (redis, rabbitmq), it's not clear to me if we need or want to set a recommended hardware level for _them_. Their performance is probably key to overall system performance.

If a user is running rabbitmq and redis locally with sensu, as I'd expect to often be the case, then I think the rabbitmq and redis requirements are equal to the sensu requirements as a whole, possibly needing an additional CPU core.

In the case of running rabbitmq and redis on other nodes, we can find a distinct set of hardware requirements for just the Ruby app.

We can probably find linear relationships between sensu events, queue message volume, and redis write volume, and that's something that should inform our estimates. Then we can also explain those relationships in our documentation, so that users can understand how the system scales. :zap: :fireworks:

palourde (approved):

Issue #310, opened 2018-03-28
Filters reference doc should describe built-in filters


cwjohnston:
We have built-in filters, extensions included in official Sensu Core packages, which should be mentioned along site native or configuration-defined filters.

cwjohnston:
- [ ] `occurrences` filter
- [ ] `check_dependencies` filter

Blue0ctober:
PR: https://github.com/sensu/sensu-docs/pull/498

PR #373, opened 2018-04-24
Moves check attribute references from plugins reference to checks refernce

### LINKED ISSUE #120: Move "occurrences" and "refresh" attribute reference documentation
_From @calebhailey on October 19, 2017 19:53_

Documentation for "occurrences" and "refresh" is currently found on the Plugins reference documentation page, rather than the Checks documentation page. The reasoning behind this was because "occurrences" and "refresh" were historically not formally supported by Sensu itself, but rather, by the `sensu-plugin` gem. This [changed in version 0.26](https://sensuapp.org/docs/0.26/overview/changelog.html#core-v0-26-0) when we began shipping the built-in "occurrences" filter. 

We even explain the original reasoning [in the Plugin reference documentation](https://sensuapp.org/docs/latest/reference/plugins.html#sensu-plugin-definition-specification) as: 

> _NOTE: plugins based on the sensu-plugin gem derive configuration from [custom check definition attributes](https://sensuapp.org/docs/latest/reference/checks.html#custom-attributes). The configuration example(s) provided above, and the “specification” provided here are for clarification and convenience only (i.e. this “specification” is just an extension of the [check definition specification](https://sensuapp.org/docs/latest/reference/checks.html#check-definition-attributes), and not a definition of a distinct Sensu primitive)._

Basically, this note is no longer true, so we can remove it and move the reference docs to the check definition specification section. 

_Copied from original issue: sensu/sensu-docs#640_

asachs01:
## Description
Moves the following out of https://docs.sensu.io/plugins/latest/reference/#check-definition-attributes to 
https://docs.sensu.io/sensu-core/latest/reference/checks/#check-attributes

```
occurrences  | 
-------------|------
description  | The number of event occurrences that must occur before an event is handled for the check.
required     | false
type         | Integer
default      | `1`
example      | {{< highlight shell >}}"occurrences": 3{{< /highlight >}}

refresh      | 
-------------|------
description  | Time in seconds until the event occurrence count is considered reset for the purpose of counting `occurrences`, to allow an event for the check to be handled again. For example, a check with a refresh of `1800` will have its events (recurrences) handled every 30 minutes, to remind users of the issue.
required     | false
type         | Integer
default      | `1800`
example      | {{< highlight shell >}}"refresh": 3600{{< /highlight >}}

dependencies | 
-------------|------
description  | An array of check dependencies. Events for the check will not be handled if events exist for one or more of the check dependencies. A check dependency can be a check executed by the same Sensu client (eg. `check_app`), or a client/check pair (eg.`db-01/check_mysql`).
required     | false
type         | Array
example      | {{< highlight shell >}}"dependencies": [
  "check_app",
  "db-01/check_mysql"
]
{{< /highlight >}}

notification | 
-------------|------
description  | The notification message used for events created by the check, instead of the commonly used check output. This attribute is used by most notification event handlers that use the sensu-plugin library.
required     | false
type         | String
example      | {{< highlight shell >}}"notification": "the shopping cart application is not responding to requests"{{< /highlight >}}
```

## Motivation and Context
Fixes #120 

## How Has This Been Tested?
Tested locally using git hooks w/ yarn

## Types of changes

- [ ] New doc/guide
- [x] Fixing errata

## Checklist:

- [x] All tests have passed.


cwjohnston:
I've suggested we make this change because the `occurrences` and `check_dependencies` filter extensions are now part of Sensu Core, and therefore these attributes have be come part of the Sensu Core check definition specification, acted on when a handler is configured with the relevant filter.

portertech:
@cwjohnston your reasoning is sound, these attributes should be documented in the Core check definition specification. My only question is, should a copy still remain under plugin reference?

majormoses (commented):
Overall I think this looks good. The only issue I see is that this does not convey that you need to apply the `occurrences` filter to the handlers to leverage this.

majormoses:
needs rebase

cwjohnston:
@majormoses good point. I think we should add a note indicating the same, and link it to the filter reference once we close https://github.com/sensu/sensu-docs/issues/310

majormoses:
> should a copy still remain under plugin reference

Hmm I think maybe we just have a little blurb and link it back to the main documentation

cwjohnston (approved):

cwjohnston:
@majormoses i added notes to those attributes.

> Hmm I think maybe we just have a little blurb and link it back to the main documentation

Perhaps the `Sensu plugin definition specification` section of the plugin reference page needs to be rewritten to describe plugin configuration conventions with examples of popular handler plugins that look at sensu json configs for their settings?  Or should we remove that section all together at this point?

cwjohnston:
@majormoses if you prefer we can leave the attributes intact in the plugins section for later refactoring

majormoses:
I am fine with either of the following:
- leave it alone and add the blurb about the need to include the filter on any extensions you expect it to work on
- remove and link to the main documentation

cwjohnston:
> remove and link to the main documentation

@asachs01 will you please add a link to the reference documentation for this change and we can ship it? :shipit:

cwjohnston:
My mistake, this change set already adds a link to the new location for these attributes. 😅

PR #389, opened 2018-05-01
Guides review

echlebek:
I've added several commits, most of which address correctness issues in the docs. I've also made a few minor changes to the English in the maintenance planning doc.

## Motivation and Context
Make sure the guides are correct before the beta release.

## How Has This Been Tested?
Manual testing.

## Types of changes

- [x] Fixing errata

## Checklist:
- [x] All tests have passed.


mercul3s (approved):
Ship it

content/sensu-core/2.0/guides/reduce-alert-fatigue.md
```diff
@@ -36,10 +36,14 @@ new events (where the event's `occurrences` is equal to `1`) or hourly events
 (so every hour after the first occurrence, calculated with the check's
 `interval` and the event's `occurrences`).
 
+Note that unlike in Sensu 1.x, events in Sensu 2.x are handled regardless of
+check execution status; even successful check events are passed through the
+pipeline. Therefore, it's necessary to add a clause for non-zero status.
```
palourde, line 41: 👍 👍 👍

content/sensu-core/2.0/guides/reduce-alert-fatigue.md
```diff
@@ -36,10 +36,14 @@ new events (where the event's `occurrences` is equal to `1`) or hourly events
...
+
 {{< highlight shell >}}
 sensuctl filter create hourly \
   --action allow \
-  --statements "event.Check.Occurrences == 1 || event.Check.Occurrences % (3600 / event.Check.Interval) == 0"
+  --statements "event.Check.Status != 0 && (event.Check.Occurrences == 1 || event.Check.Occurrences % (3600 / event.Check.Interval) == 0)"
```
palourde: The problem I see with that new condition is that events that just resolved will not be handled, which could be really problematic with a handler such as PagerDuty. Ideally, we only want the first occurrence of an event with `status == 0` to go through the pipeline, but I was thinking maybe it would require a new status, `resolving` for example, for the  `event.Check.State` field, which would only be set for the first occurrence of a successful check. Otherwise, I'm not sure how we can easily do that on this filter without adding a ton of conditions. Thoughts?

echlebek: I agree that we need a "Resolved" field for cases like this. It would be set true after the first instance of the status going from non-zero to zero, false for all other conditions.

palourde: So now that we are aware of the existence of the `is_incident` built-in filter, we should probably revert that change and instruct the users to assign both the `is_incident` and `hourly` filter to their handler!

PR #405, opened 2018-05-07
Clarification around not_silenced built-in filter

palourde:
Make a distinction between silencing and not handling silenced events.


content/sensu-core/2.0/guides/plan-a-maintenance.md
```diff
@@ -75,6 +76,10 @@ $ sensuctl event list
   scotch.local    keepalive                     0          true     2018-03-16 13:22:16 -0400 EDT
 {{< /highlight >}}
 
+_WARNING: A silenced event does not automatically means that it will not be
```
mercul3s: s/means/mean

mercul3s: Would also suggest something like "By default, a silenced event will be handled **unless** the handler uses the `not_silenced` filter to discard silenced events" to avoid the double negative sentence here.

mercul3s (changes requested):

palourde:
@mercul3s Ready for review!

mercul3s (approved):
🍕 🌮 🍨 🍔 🌯

PR #408, opened 2018-05-07
Clarification around is_incident built-in filter

palourde:
Make it clear that all events are handled, by default, regardless of their status.


mercul3s (approved):
🍕 🌮 🍨 🍔 🌯

PR #470, opened 2018-06-07
Update aggregates reference docs to include aggregates check definition

Evesy:
## Description
The check attributes listed in the aggregates reference section do not include `aggregates`, the check attribute to pass an array of aggregates, instead of using `aggregate` which only accepts a single aggregate string.

`default: false` has also been removed from the `aggregate` check definition reference as this is not a boolean option with a default value

## Motivation and Context
Provides full list of aggregate check attributes within the aggregates reference section; bringing in line with the aggregate related options outlined in the `checks` reference documentation.

## How Has This Been Tested?
`hugo build`

## Types of changes
- [ ] New doc/guide
- [x] Fixing errata
- [x] Update (Add missing or refresh existing content)

## Checklist:

- [x] All tests have passed.


apaskulin (approved):
Thanks for adding this!

PR #516, opened 2018-06-22
fixed missing step and typo

danhimalplanet:
need to install uchiwa package to use it

you start a service with "start", not "stat"


apaskulin (approved):
Thanks for fixing this!

PR #523, opened 2018-06-27
Update ansible gem docs

jaredledvina:
* Switch to native YAML format for easier reading
* We need to explicitly pass `user_install: no`


majormoses (approved):
LGTM

Issue #228, opened 2018-03-02
Guide Wanted: Scaling Sensu


cwjohnston:
We've been promising that a guide on scaling Sensu is ["coming soon"](https://sensuapp.org/docs/latest/installation/summary.html#scaling-sensu) for a long time now. We should make good on this.

I've found an older guide to Sensu scalability in the legacy documentation, e.g.: https://github.com/sensu/sensu-docs/blob/master/legacy/0.16/scaling_strategies.md

I think we ought to use this document as the basis of an improved guide.

asachs01:
Base on the team conversation today, I'm rescoping the original PR in #324 so that it's not so broad. That said, there will be several new docs:

- [x] Scaling overview (will include the scaling strategies discussed in the article ^ ) (covered by #565)
- [ ] Knowing when to scale Sensu components
- [ ] Scaling Sensu at Multiple Sites: Isolated Clusters Aggregated by Uchiwa 
- [ ] Scaling Sensu at Multiple Sites: Centralized Sensu and Distributed RabbitMQ
- [ ] Scaling Sensu at Multiple Sites: Centralized Sensu and Directly Connected Clients

Each of the last 3 guides will feature more in-depth discussions of their respective topologies

PR #565, opened 2018-07-11
Core/add/scaling sensu guide

asachs01:
## Description
Scoped down to address the first bit of #228 . 

## Motivation and Context
Provides a brief overview of scaling Sensu components and scaling strategies.

## How Has This Been Tested?
`yarn server`

## Types of changes

- [x] New doc/guide
- [ ] Fixing errata

## Checklist:

- [x] All tests have passed.


content/sensu-core/0.29/guides/scaling-overview.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### Pros
+
+* WAN instability does *not* lead to flapping Sensu checks
+* Sensu operation continues un-interrupted during a WAN outage
+* The overall architecture is easier to understand and troubleshoot
+
+#### Cons
+
+* WAN outages mean a whole Datacenter can go dark and not set off alerts (cross-datacenter checks are therefor essential)
```
rgeniesse: Maybe I am not following this correctly, but you list a Con here where a `Datacenter can go dark and not set off alerts`, but a Pro above is `Sensu operation continues un-interrupted during a WAN outage `. These seem to contradict each other a bit.

Feels like the con should be that a whole Datacenter can go dark in Uchiwa, giving no dashboard visibility. That said, depending on the outage, it could mean that the Datacenter can't alert on itself either, but it is possible that just your centralized Uchiwa dashboard can't reach the API of the other data center(s), but those other datacenter(s) can still reach the internet, which would allow events to still be handled and be sent out.

apaskulin (commented):
This guide does a great job laying out the different considerations involved in scaling Sensu. I've added some formatting suggestions to https://github.com/sensu/sensu-docs/commit/39608ebed0413bd5f38846d4406d5ee8c8fbc44c

I'm curious to see the content for Scaling Sensu Components and hear your thoughts on whether that content would be better situated in the troubleshooting guide or in the reference docs.

content/sensu-core/0.29/guides/scaling-overview.md
asachs01: Good point. That's pulled from the original doc here https://github.com/sensu/sensu-docs-legacy/blob/master/legacy/0.16/scaling_strategies.md , so maybe we're both lacking a bit of context? Not sure if @cwjohnston would have any additional context for it, or could interpret.

asachs01:
@apaskulin the way that I'm envisioning that content flowing would be less of a troubleshooting-style doc, or reference doc, and more of a "here are some behaviors that when observed in your environment, should clue you in that it's time to scale", if that make sense. Although, we have those reference docs already, so it may be less effort to incorporate those as blurbs in the reference doc?

asachs01:
@apaskulin also thank you for the changes! Pulled those in.

apaskulin (approved):
Looks good to me!

Issue #581, opened 2018-07-16
Review sidebar ordering


apaskulin:
Ordering of links in the navigation sidebar is controlled by the `weight` parameter in the page frontmatter.

To do:
- Review page weights so that sections and pages are ordered logically
- Check for erroneous sidebar items (ex: https://docs.sensu.io/sensu-core/1.4/quick-start/)
- Check for conflicting or duplicated weight values

richburroughs:
Taking the bug label off. This sounds like a content issue. @apaskulin is this something you can handle. Or do you need help?

apaskulin:
Thanks, I can take care of this, @richburroughs.

apaskulin:
Closed by #1121, #670, and #648

PR #621, opened 2018-07-30
Order menu sections by weight

palourde:
## Description

This PR adds sorting by weight for menu entries.

Right now, entries for a given menu (e.g. sensu-core-1.4) are alphabetically sorted, even if a weight is specified in a `_index.md` file, for example:

```
[...]
menu:
  sensu-core-1.4:
    identifier: getting-started
    weight: 100
```

## Motivation and Context
Related to https://github.com/sensu/sensu-docs/issues/581.

## How Has This Been Tested?
Locally tested with 2.0 documentation.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New doc/guide
- [ ] Fixing errata

## Checklist:

- [x] All tests have passed.

### DIFF
--- a/layouts/partials/nav.html
+++ b/layouts/partials/nav.html
@@ -31,7 +31,7 @@
           <li><a class="menu-title" href="/{{ $url }}/{{ $.Params.version }}">{{ $niceName }}</a><ul class="toc-menu-container">
 
           <!-- Loop through the menus pages -->
-          {{ range $y, $x := $v }}
+          {{ range $y, $x := sort $v "Weight" }}
 
           <!-- Determine if one the children of this menu is active so we can apply the proper styling -->
             {{ $isCurrent := ($currentNode.HasMenuCurrent $k $x) }}


apaskulin:
@palourde This seems like it works, but I'm not able to reproduce the initial issue. It seems like menu ordering by weight currently does work. Can you describe where you're seeing an issue?

palourde:
@apaskulin Menu ordering does work for pages within a section, but it does not work but sections themselves.

For example, Sensu Core 1.4 menu appears like this:

screen shot 2018-07-30 at 12 21 56 pm

Let's say I wanted to move the _Quick Start_ section to the top and the _API_ one at the bottom, I should be able to use the `weight` attribute in the section files, however it does not work.

The problem right now is that the `weight` attributes defined in the `_index.md` files have no effect because the `nav.html` template does not take in consideration the section's weight when ranging over them. This PR addresses this specific problem.

Additionally, some of our `_index.md` files are not properly configured; they should specify their `identifier` instead of referring to themselves as their parent. For example, in the file `content/sensu-core/1.4/api/_index.md`, instead of having this

```
[...]
 menu:
   sensu-core-1.4:
     parent: api
```

We should do this:
```
# We should do this instead
[...]
 menu:
   sensu-core-1.4:
     identifier: api
```

Once these two problems are fixed for all sections, the menu will properly sort the sections:

screen shot 2018-07-30 at 12 53 44 pm

apaskulin (approved):
This is awesome! Thanks so much for doing this 🎉 🌟 Can't wait to start ordering these sections. I'll open an issue to get the frontmatter updated on the index pages.

Issue #577, opened 2018-07-16
2.0 Glossary


apaskulin:
Create a glossary of terms for Sensu Core 2.0 under `2.0/getting-started` with the following terms:

- [x] agent
- [x] backend
- [x] sensuctl
- [x] asset
- [x] check
- [x] entity
- [x] event
- [x] proxy entity
- [x] filter
- [x] handler
- [x] hook
- [x] mutator
- [x] plugin
- [x] RBAC
- [ ] silencing
- [ ] subdued
- [ ] token

Reference: #568, #433

PR #638, opened 2018-08-01
Add glossary for Sensu 2.0

apaskulin:
## Description

This PR adds a glossary of basic terms for Sensu 2.0. Feedback very welcome!

## Motivation and Context

Partially satisfies #577

## How Has This Been Tested?

Locally via yarn

## Types of changes

- [x] New doc/guide
- [ ] Fixing errata

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] All tests have passed.


content/sensu-core/2.0/getting-started/glossary.md
```diff
@@ -0,0 +1,68 @@
...
+---
+
+#### Agent
+A lightweight client that runs on the infrastructure components you want to monitor. Agents self-register with the backend, send keepalive messages, and execute monitoring checks. Each agent belongs to one or more subscriptions that determine which checks the agent runs. An agent can run checks on the entity it’s installed on or by connecting to a remote proxy entity. [Read more.][1]
+
+#### Backend
+A flexible, scalable monitoring event pipeline. The backend processes event data using filters, mutators, and handlers. It maintains configuration files, stores recent event data, and schedules monitoring checks. You can interact with the backend using the API, command line, and dashboard interfaces. [Read more.][2]
+
+#### Check
+A script run by the agent to determine the state of a system component or collect metrics. Checks specify an interval for execution, one or more subscriptions, and one or more handlers that will process the resulting event data. Once run, a check produces a check result that includes the output of the command and an exit status code. [Read more.][3]
```
rgeniesse: So just a general thing from a 1.x perspective: 1.x has check definitions and the check executable that those definitions run, which as a whole is usually just called a "check". This definition appears to be talking a little about both as it starts out saying a check is "A script", but then it has other Sensu terms talked about. 

Entirely possible that is the right way to think of things in 2.x, but it is definitely something I like to distinguish in Sensu 1.x as it helps with the learning process.

content/sensu-core/2.0/getting-started/glossary.md
```diff
@@ -0,0 +1,68 @@
...
+
+#### Check asset
+A script that a check can specify as a dependency. Assets must be a tar archive (optionally gzipped) with scripts or executables within a bin folder. At runtime, the agent installs required assets using the specified URL. Assets let you manage runtime dependencies without using configuration management tools. [Read more.][4]
+
+#### Check hook
+A command executed by the agent in response to a check result, before creating a monitoring event. Hooks create context-rich events by gathering related information based on the exit status code. [Read more.][5]
+
+#### Check plugin
+Checks require a script that the agent runs on the corresponding entity when executing the check. You can write your own check scripts in Go, Ruby, Python, and more, or use one of over 100 check plugins shared by the Sensu community. [Read more.][6]
```
rgeniesse: Sorta goes back to my comment above where a Check Plugin is the executable, but a Check would be the whole definition including the Check Plugin. Nothing to change here, just wanted to show that relationship further.

content/sensu-core/2.0/getting-started/glossary.md
```diff
@@ -0,0 +1,68 @@
...
+Infrastructure components that you want to monitor. Each entity runs an agent that executes checks and creates events. Events can be tied to the entity where the agent runs or a proxy entity that the agent checks remotely. [Read more.][7]
+
+#### Event
+A representation of the state of an infrastructure component at a point in time, used by the backend to power the monitoring event pipeline. Event data includes the result of the check or metric (or both), the executing agent, and a timestamp. [Read more.][8]
+
+#### Filter
+Logical statements that handlers evaluate before processing monitoring events. Filters can instruct handlers to allow or deny matching events based on day, time, organization, environment, or any attribute in the event data. [Read more.][9]
+
+#### Handler
+A script run by the backend to act on events. Handlers can send monitoring event data to a script (or handler plugin), a TCP socket, or a UDP socket. [Read more.][10]
```
rgeniesse: `ran` or maybe `that is run`? I don't know, reads weird to me for some reason. Maybe it is fine. Ignore if you feel it is, :-).

content/sensu-core/2.0/getting-started/glossary.md
apaskulin: Would this be better?

>A script run by the agent to determine the state of a system component or collect metrics. Check definitions specify the location of the script, an interval for execution, one or more subscriptions, and one or more handlers that will process the resulting event data. Once run, a check produces a check result that includes the output of the command and an exit status code.

rgeniesse: I think I am being difficult here, or something. Or taking words too literally. This is possibly entirely fine the way it is.

My overall idea here was to ensure we distinguish between a check (the thing defines what command a check runs, the interval, etc.) and a check plugin (the thing that gets ran). Calling a check (I read that as a check definition a script, to me, isn't right in 1.x; it might be valid in 2.x.

Let's just leave it alone unless someone else feels the same way.

mercul3s (approved):
🍕 🌮 🍨 🍔 🌯

PR #690, opened 2018-08-23
Update transport reference with more explicit caveats for redis transport

cwjohnston:
## Description

Updates the list of cons and additional copy describing the Redis transport

## Motivation and Context

Redis transport has known bugs which prevent it from being fully supported. The documentation currently does not make this clear enough to help evaluators or implementers avoid known problems with this transport.

I’m not sure why, but compared to older versions of this document, it seems we lost some of the necessary clarity in our site migration project. See https://github.com/sensu/sensu-docs-legacy/blob/master/docs/1.2/reference/transport.md for comparison.

### DIFF
--- a/content/sensu-core/0.29/reference/transport.md
+++ b/content/sensu-core/0.29/reference/transport.md
@@ -54,11 +54,11 @@ recommended solution for running Sensu in production environments.
 
 ### The Redis Transport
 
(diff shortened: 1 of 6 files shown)


asachs01 (approved):
Lgtm

PR #590, opened 2018-07-19
New messaging for Core 1.x

apaskulin:
## Description

- First draft of new diagrams and messaging for Sensu Core 1.x landing page.
- Sensu 2.0 landing page content
Looking for all the feedback.

See the previews:
- https://sensu-docs-stage.herokuapp.com/sensu-core/1.4/ 
- https://sensu-docs-stage.herokuapp.com/sensu-core/2.0/

## Motivation and Context

Closes #584 

## How Has This Been Tested?

Locally via yarn

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New doc/guide
- [ ] Fixing errata
- [x] Update (Add missing or refresh existing content)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] All tests have passed.

PR #697, opened 2018-08-29
Update messaging to use "ephemeral"

apaskulin:
## Description

After talking to several folks, I'd like to change "containerized" to "ephemeral" in the central messaging for Core and Enterprise. I think this term is closer to what we're trying to describe here, and it's a term that people running these types of infrastructures identify with.

## Motivation and Context

Follow up to #590


rgeniesse (approved):
👍 👍

PR #705, opened 2018-08-31
Recommend version pinning for plugins

apaskulin:
The [community plugin docs](https://github.com/sensu-plugins/community/blob/master/best_practices/production_deployments/plugins/PINNING_VERSIONS.md) recommends pinning versions when installing plugins via CLI or config management tools. We should add this recommendation to the Installing Plugins page within Sensu Core.

## Description

- Fixes typos in existing doc
- Adds note recommending version pinning with examples
- Style fixes to make content more modular and use title caps for plugin names

To do: Copy to all versions post-review

## Motivation and Context

In response to feedback from @majormoses via Slack.
Since the docs page doesn't mention configuration management, I opted to include just the CLI commands for pinning versions without linking to the [community plugin docs](https://github.com/sensu-plugins/community/blob/master/best_practices/production_deployments/plugins/PINNING_VERSIONS.md).

## Review Instructions
Please review https://github.com/sensu/sensu-docs/commit/2350ed4465dd33ae1989ac15e4d7daaa7250b05f as I wasn't able to test on Windows.


content/sensu-core/1.5/installation/installing-plugins.md
```diff
@@ -46,19 +49,30 @@ Successfully installed sensu-plugins-disk-checks-3.1.0
...
+sudo sensu-install -p 'sensu-plugins-disk-checks:3.1.0'
+{{< /highlight >}}
+
 ### Removing Plugins
 
 If you find that you need to remove a plugin, you can use the embedded `gem` command to remove a plugin. See the example below:
 
 {{< highlight shell >}}
-/opt/sensu/embedded/bin/gem uninstall sensu-plugins-disk-checks{{< /highlight >}}
+sudo /opt/sensu/embedded/bin/gem uninstall sensu-plugins-disk-checks{{< /highlight >}}
```
majormoses: Depending on your setup `sudo` may not be necessary.

content/sensu-core/1.5/installation/installing-plugins.md
```diff
@@ -46,19 +49,30 @@ Successfully installed sensu-plugins-disk-checks-3.1.0
 1 gem installed
 [SENSU-INSTALL] successfully installed Sensu plugins: ["sensu-plugins-disk-checks"]{{< /highlight >}}
 
+### Pinning Versions
+
+We strongly recommend pinning the versions of any plugins installed in production.
+To install the [Sensu Disk Checks Plugin][3] and pin the version, use:
+
+{{< highlight shell >}}
+sudo sensu-install -p 'sensu-plugins-disk-checks:3.1.0'
```
majormoses: Depending on your setup `sudo` may not be necessary.

content/sensu-core/1.5/installation/installing-plugins.md
```diff
@@ -46,19 +49,30 @@ Successfully installed sensu-plugins-disk-checks-3.1.0
...
+{{< /highlight >}}
+
```
majormoses, line 58: I think there should link to this guide: https://github.com/sensu-plugins/community/blob/master/best_practices/production_deployments/plugins/PINNING_VERSIONS.md for further reading as it goes into more depth on the rationale and examples with `cli using gem directly`, `sensu-install`, `chef`, `puppet`, `ansible`, and `saltstack`.

apaskulin: Since this page doesn't mention configuration management, I'm going to hold off linking to this page for now. #712 is tracking more general efforts around incorporating these awesome best practices into the docs.

majormoses: Not sure why that matters (config management), seems like duplicate effort but OK.

content/sensu-core/0.29/installation/installing-plugins.md
```diff
@@ -46,6 +49,15 @@ Successfully installed sensu-plugins-disk-checks-3.1.0
 1 gem installed
 [SENSU-INSTALL] successfully installed Sensu plugins: ["sensu-plugins-disk-checks"]{{< /highlight >}}
 
+### Pinning Versions
+
+We strongly recommend pinning the versions of any plugins installed in production.
```
cwjohnston: I think we need to do more here to describe what version pinning is. To me "version pinning" is a concept that involves both installing a specific version of a package, and using a mechanism to enforce that version. The sensu-plugin tool enables the former but does nothing to cover the latter requirement.

cwjohnston (changes requested):

content/sensu-core/0.29/installation/installing-plugins.md
majormoses: That makes sense to me, we can add a bit about that in the main document about pinning and link here.

majormoses: @cwjohnston just to clarify what you mean in the case of os package managers such `apt`, `yum`, etc they have a mechanism to create a lock/hold which would prevent someone doing a generic update from installing it. In the case of ruby gem installs there is no such global guarantee though in your automated process be it `chef`, `puppet`, `ansible`, etc would ensure that the version to be installed (defined by strict or more loose constraints) will ensure that the version is installed but can't prevent someone with enough access from going to the CLI and manually installing it. The closest mechanism that exists for that is `bundler` but thats really for specific apps and other apps can install newer versions into a shared context if given access.

cwjohnston (approved):

content/sensu-core/1.5/installation/installing-plugins.md
apaskulin: Updated to link to the best practices doc!

PR #711, opened 2018-09-01
Add note that handle_when used inside a handler uses defaults

### LINKED ISSUE #571: Update handle_when Enterprise filter documentation
## This is an issue with:

## Expected Behavior
Documentation clearly states when you can use `handle_when` attributes.

## Current Behavior
Documentation does not have a statement that `handle_when` attributes are not configurable when the filter is used with a handler.

## Possible Solution
Update documentation. Possible this applies to other built in filters.

rgeniesse:
## Description
Closes #571 

## Motivation and Context
It is not clear that using this filter within a handler makes it so you can not change the attributes that handler_when offers from their default values.

## Review Instructions
Run locally with `yarn` to view changes.


content/sensu-enterprise/2.6/filters/handle-when.md
```diff
@@ -59,6 +59,8 @@ enterprise filter will have events filtered unless these conditions are met.
 The following is an example of how to apply the `handle_when` enterprise filter
 to a standard Sensu `pipe` handler.
 
+_NOTE: The `handle_when` filter, when used inside a handler, does not allow the `occurrences` or `reset` attributes to be changed from their defaults._
```
cwjohnston: In the spirit of using clear and affirmative language, I think we should reframe this note. How about something like this:

> The default values for `occurrences` and `reset` attributes can be overridden by specifying custom values in your check definitions as shown above. Specifying values for `handle_when` attributes in a handler definition will have no effect.

cwjohnston (changes requested):

cwjohnston (approved):

PR #717, opened 2018-09-05
Bug fix: Navigation link

apaskulin:
## Description

Since introducing relative URLs, the site no longer has a BaseURL. The link to the navigation page in the sidebar uses the BaseURL to create the link, so it is currently not working in prod or staging. However when running the site locally, the BaseURL does get set to `localhost:1313`. This PR hardcodes the link to the docs site homepage, although it would be better to find a solution that would work for staging as well.

## Motivation and Context

Bug causing "Navigation" link in sidebar to not work


asachs01 (approved):
LGTM :+1:

PR #724, opened 2018-09-11
Add JSON validator for .md files inside the content directory

rgeniesse:
## Description
Validates JSON within tags `{{< highlight json >}} {{< /highlight >}}` using Python's json library.

## Motivation and Context
Ensures future JSON blocks are valid inside the `content` directory.


.travis.yml
```diff
@@ -1,13 +1,25 @@
...
-      - python-pygments
+      addons:
+        apt:
+          packages:
+            - python-pygments
 
-before_install:
-  - export PATH="$TRAVIS_BUILD_DIR/bin:$PATH"
+      before_install:
+        - export PATH="$TRAVIS_BUILD_DIR/bin:$PATH"
```
apaskulin, line 13: Are these spacing changes intentional?

rgeniesse: Yes. YAML depends on spacing to dictate what an object lives under. So as I added a second language here, I needed to move these over they would be under `matrix` and `include`. 

The spacing that doesn't matter are the new lines between objects, but I just followed what was here to be consistent.

apaskulin: Perfect. Thanks!

apaskulin (approved):
This is awesome! So happy to have this added to the project. I wasn't able to test this, but I'm +1 to merging and iterating. :tada: ⭐️

PR #732, opened 2018-09-12
Add extended attributes and environment variables to Sensu 2.0 checks reference

apaskulin:
## Description

This PR adds two missing attributes to the checks reference docs: `extended_attributes` and `env_vars`

## Motivation and Context

Checking doc for completeness against https://github.com/sensu/sensu-go/blob/master/types/check.proto


content/sensu-core/2.0/reference/checks.md
```diff
@@ -285,7 +285,10 @@ example      | {{< highlight shell >}}"round_robin": false{{< /highlight >}}
 
 |extended_attributes|      |
 -------------|------
-description  | Coming soon.
+description  | Custom attributes to include with the event data, which can be queried like regular attributes.
+required     | false
+type         | JSON object
```
echlebek: The type here should say "Serialized JSON object" - users who are creating checks with `sensuctl create` could be tripped up by that.

content/sensu-core/2.0/reference/checks.md
```diff
@@ -285,7 +285,10 @@ example      | {{< highlight shell >}}"round_robin": false{{< /highlight >}}
...
+example      | {{< highlight shell >}}{"team":"ops"}{{< /highlight >}}
```
echlebek: This should be modified as follows:
```
example      | {{< highlight shell >}}"{\"team\":\"ops\"}"{{< /highlight >}}
```

echlebek (changes requested):
Nice work finding these holes in the documentation! :heart: 

We've had some users get tripped up over extended attributes, so we should be careful to explain that they are to be specified as serialized JSON.

We're considering reworking extended attributes, so it might be better to not mention them at all. @grepory do you have any comment on this?

grepory:
The documentation should continue to say that they should be specified as a JSON-serialized hash. The internals of how we support them and how they are accessed in filters/plugins will change, but I don't believe the UX will fundamentally change.

apaskulin:
Thanks, @echlebek! I added those corrections to the extended attributes description in  https://github.com/sensu/sensu-docs/pull/732/commits/7c96ea276e356fa94e161db4a934e37216409b1f

cwjohnston (approved):

PR #734, opened 2018-09-13
Increase h2 font size

apaskulin:
## Description

Currently, the h2 and h3 font styles are quite close in size, 24px and 22px respectively. This makes it difficult to visually differentiate between h2 and h3 headings, which is essential for page navigation. This PR increases the h2 size to 27px.

Before (left) and After (right):

screen shot 2018-09-13 at 8 36 34 am

## Motivation and Context

Resolves issue with page styling


cwjohnston (approved):
I started noticing the same issue while working on some new content today. Nice work! 👍

PR #752, opened 2018-09-20
Add note about an interval timer's initial offset

nikkictl:
## Description
Adds a note to `Check Scheduling` regarding interval and cron schedules. Briefly describes an interval timer's initial offset.

## Motivation and Context
Helps solve some confusion discovered in https://github.com/sensu/sensu-go/issues/2053 and addressed in https://github.com/sensu/sensu-go/pull/2087.


apaskulin (approved):
Yay for clarification notes! 🎉 

It would be great to include a connection to the user experience, something like: This helps to balance the load of both the backend and the agent and may result in a delay before initial check execution.

PR #765, opened 2018-09-27
Fallback to local fontawesome js in case cdn down

mercul3s:
This is hardcoded now until I can get the fallback src url interpolated
properly, but should resolve our issues with maxcdn being down and
causing the side to hang attempting to load fontawesome.

Signed-off-by: Mercedes Coyle <mercedes@sensu.io>


amdprophet (approved):
Looks good and works on staging. 👍

Issue #764, opened 2018-09-26
Site hanging due to Fontawesome.js unavailability


richburroughs:
## This is an issue with:

- [X] Bug (site functionality or styling)
- [ ] Errata (fix needed for doc content)
- [ ] New content (guide wanted)
- [ ] Update (Add missing or refresh existing content)

## Expected Behavior

Site should be loading.

## Current Behavior

Site is hanging due to fontawesome's CDN being unavailable. https://twitter.com/fontawesome/status/1045070553734950912

## Possible Solution

We could possibly host the file locally, or come up with a different solution.

## Steps to Reproduce (for bugs)

1.
2.
3.

## Context

* Browser(s) and version(s): 
* OS:

mercul3s:
This needs a fallback to a locally hosted fontawesome.js if the cdn is unavailable.

mercul3s:
A quick and dirty fix has been deployed with this pr: https://github.com/sensu/sensu-docs/pull/765
I still need to ensure the actual fallback mechanism works, will do that later this week.

mercul3s:
I have a working mechanism to do fallback, so I'm going to go ahead and vendor any other cdn sourced assets as well.

richburroughs:
@mercul3s Please do. Thank you :)

PR #775, opened 2018-10-02
Add real user monitoring

### LINKED ISSUE #770: Add Pingdom RUM
Enable real user monitoring via Pingdom for the docs site.

To do:

- [ ] Add the required script tag
- [ ] Deploy to staging and check for performance impacts

apaskulin:
## Description

Adds the real-user monitoring script from Pingdom

## Motivation and Context

As a follow-up to the the site outage described by #764, this PR makes our Pingdom monitoring setup more robust by adding real-user monitoring.

Closes #770  

## Review Instructions

Live on staging https://sensu-docs-stage.herokuapp.com/

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -69,6 +69,7 @@
       crossorigin="anonymous"></script>
 
 	<script src="https://use.fontawesome.com/ea9395e21e.js"></script>
+	<script src="//rum-static.pingdom.net/pa-5bacf02ccea07b00160001cc.js" async></script>
 
   <!-- Local fallback for assets in case of cdn failure -->
 	<script type="text/javascript">


Blue0ctober (approved):
👍

rgeniesse:
We'll want to make sure that adding this won't cause the documentation site to become un-responsive if that added resource becomes unavailable. The `async` attribute may overall fix that based on the definition of it:

https://www.w3schools.com/tags/att_script_async.asp

But probably worth a test in staging by giving a bad script name. This may not cover all cases though, as behavior might change if the site is up but script is missing vs the site being down. I say this lightly though, as it isn't based on any knowledge, just the fact that those scenarios of down are technically different (not found resource vs network outage).

PR #801, opened 2018-10-15
Add docs for check delete endpoint

### LINKED ISSUE #189: Write up what happens when you rename or delete a check or client
Something that catches a lot of people by surprise is when they rename or remove a check and it still shows up in uchiwa. This is because unless the check has a TTL configured sensu keeps it around in redis and uchiwa aside from its functionality integrating into the api changing state is essentially a visual representation of the state. As such the check will remain even after removing the file and restarting api, server, and client. This can typically be solved via two methods:
1. clean out the keys for that check on the clients
2. delete the client and let it re-register

The basic steps as I understand them is to locate the keys in redis:
```
$ redis-cli
127.0.0.1:6379> keys '*:ip-10-38-152-75.us-west-2.compute.internal:disk'
1) "history:ip-10-38-152-75.us-west-2.compute.internal:disk"
2) "result:ip-10-38-152-75.us-west-2.compute.internal:disk"
```

Now that we know which keys to delete we can either delete them one by one or use several ways of dealing with looping over they keys and deleting them.

Nothing quite like good old reliable `xargs`:
```bash
$ redis-cli KEYS "prefix:*" | xargs redis-cli DEL
```
Another option if you have a new enough version of redis you can execute a lua script to do the same thing:
```
$ redis-cli
EVAL "return redis.call('del', unpack(redis.call('keys', ARGV[1])))" 0 prefix:*
```

In both of these cases update the prefix to match your check name and node name. You want to minimally remove the result and history keys.

For deleting a client:
- In Uchiwa there is an x in the top right corner of the client
- use the api

Example API request without auth:
```
curl -XDELETE --silent -L https://$SENU_API_HOST:4567/clients/$CLIENT_NAME
```

### LINKED ISSUE #797: Document Check result deletion API endpoint
Reference: https://github.com/sensu/sensu/pull/1909

apaskulin:
## Motivation and Context

Closes #797 
Closes #189

## To do

- [x] Check renaming flow
- [ ] Client delete/rename for all versions
- [x] `curl -s -i -X DELETE http://127.0.0.1:4567/checks/non-existent-check` returns 202 instead of 404

### DIFF
--- a/content/sensu-core/1.6/api/checks.md
+++ b/content/sensu-core/1.6/api/checks.md
@@ -166,9 +166,9 @@ Server: thin
 
 /clients/:client (DELETE) | 
 --------------------------|------
-description               | Removes check results and check history for a specified check across all clients. This endpoint does not modify any configuration files.
+description               | Asynchronously removes check results and check history for a specified check across all clients. This endpoint does not modify any configuration files.
 example url               | http://hostname:4567/checks/sensu_website
-response codes            | <ul><li>**Success**: 202 (Accepted)</li><li>**Missing**: 404 (Not Found)</li><li>**Error**: 500 (Internal Server Error)</li></ul>
+response codes            | <ul><li>**Success**: 202 (Accepted)</li><li>**Error**: 500 (Internal Server Error)</li></ul>
 output                  | {{< highlight shell >}}HTTP/1.1 202 Accepted
 Content-Type: application/json
 Access-Control-Allow-Origin: *
--- a/content/sensu-core/1.6/api/checks.md
+++ b/content/sensu-core/1.6/api/checks.md
@@ -143,7 +143,8 @@ and history.
 #### EXAMPLE {#checkscheck-delete-example}
 
 The following example demonstrates a request to delete a check named
-`sensu_website`, resulting in a [202 (Accepted) HTTP response code][3].
+`sensu_website`, resulting in a [202 (Accepted) HTTP response code][3] and
+and a JSON Hash containing an `issued` timestamp.
 
 {{< highlight shell >}}
 $ curl -s -i -X DELETE http://127.0.0.1:4567/checks/sensu_website
@@ -157,6 +158,8 @@ Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Au
 Content-Length: 21
 Connection: keep-alive
 Server: thin
+
+{"issued":1539627787}
 {{< /highlight >}}
 
 #### API Specification {#clientsclient-delete-specification}
@@ -175,6 +178,8 @@ Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Au
 Content-Length: 21
 Connection: keep-alive
 Server: thin
+
+{"issued":1539627787}
 {{< /highlight >}}
 
 ## The `/request` API endpoint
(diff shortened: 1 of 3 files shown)


content/sensu-core/1.6/api/checks.md
```diff
@@ -134,6 +135,53 @@ output               | {{< highlight json >}}{
...
+Access-Control-Allow-Credentials: true
+Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Authorization
+Content-Length: 21
+Connection: keep-alive
+Server: thin
+
+{"issued":1539627787}
+{{< /highlight >}}
+
+#### API Specification {#clientsclient-delete-specification}
```
cwjohnston: Anchor link seems wrong

content/sensu-core/1.6/api/checks.md
```diff
@@ -134,6 +135,53 @@ output               | {{< highlight json >}}{
...
+
+/clients/:client (DELETE) |
```
cwjohnston: API endpoint should be `/checks/:check`

content/sensu-core/1.6/reference/checks.md
```diff
@@ -347,6 +348,36 @@ outputting the last 100 lines of an error log.
...
+
+{{< highlight shell >}}
+# Delete check results and check history
+curl -s -i -X DELETE http://127.0.0.1:4567/checks/check-sensu-website
+{{< /highlight >}}
+
+### Rename a check
+
+Since checks are defined by their name, checks cannot be renamed.
+In practice, you can rename a check by [deleting it](#delete a check) and re-creating it with a new name.
```
cwjohnston: I think we're conflating check definitions, which can be renamed, with check results. Here's an alternate take:

> Because there's no direct relationship between check results in the data store and configured check definitions, the name of a stored check results can't be changed.
> In practice, you can rename a check by changing it's name in the relevant Sensu configuration file, and [deleting the results attributed to  the former name](#delete a check).

cwjohnston (changes requested):
Seems like we need couple of changes here.

content/sensu-core/1.6/reference/checks.md
```diff
@@ -347,6 +348,36 @@ outputting the last 100 lines of an error log.
 }
 {{< /highlight >}}
 
+## Managing checks
+
+### Delete a check
+
+To delete a check from Sensu, remove the configuration file for the check, then restart the Sensu server and API.
```
rgeniesse: Do we know if this endpoint supports standalone checks? I want to say no.

If they are supported, we need to call out that the Sensu client where the check definition would need restarted.
If they are not supported, we should call out that this won't work for stand alone checks either here or possibly here and/or elsewhere.

My lab is down at the moment. I'll work on bringing it up to test.

content/sensu-core/1.6/reference/clients.md
```diff
@@ -419,6 +420,18 @@ title="typically accurate within 500ms">consistency</abbr> between standalone
 checks _across an entire infrastructure_ (assuming that system clocks are
 synchronized via [NTP][23]).
 
+## Managing clients
+
+### Delete a client
+
+While shutting down a client stops all check and keepalive processes, it doesn't affect the history of the client or remove existing keepalive results.
```
rgeniesse: Wonder if for `check and keepalive processes` use `check results and keepalives` instead, as the process you are shutting down is the `sensu-client` here.

content/sensu-core/1.6/reference/checks.md
apaskulin: Interesting. My thought here was that Sensu doesn't see a difference between a new check and a renamed check, but I've updated this to use the workflow you've outlined, which is also more in line with the rename-a-client workflow.

content/sensu-core/1.6/reference/clients.md
apaskulin: Great callout! I've updated this.

cwjohnston (approved):

PR #608, opened 2018-07-26
Update Erlang and RabbitMQ RPM install docs

cwjohnston:
## Description

Removes manual commands specifying particular versions of Erlang and RabbitMQ, replacing them with references to RabbitMQ's own canonical documentation.

## Motivation and Context

Closes #450 

## How Has This Been Tested?

yarn server

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New doc/guide
- [x] Fixing errata
- [ ] Update (Add missing or refresh existing content)

## Checklist:
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] All tests have passed.

PR #836, opened 2018-10-29
Updates install instructions

### LINKED ISSUE #835: Update Debian Install Docs
## This is an issue with:

- [x] Errata (fix needed for doc content)

## Expected Behavior
We should be pointing at RMQ's install docs instead of maintaining our own.

## Current Behavior
We're on the hook for updating deb package versions

## Possible Solution
Punt to RMQ http://www.rabbitmq.com/install-debian.html#apt

asachs01:
## Description
Closes #835 

## Motivation and Context
Takes Sensu off the hook for maintaining install docs/updating on every version

### DIFF
--- a/content/sensu-core/0.29/installation/install-rabbitmq-on-ubuntu-debian.md
+++ b/content/sensu-core/0.29/installation/install-rabbitmq-on-ubuntu-debian.md
@@ -22,6 +22,8 @@ Although RabbitMQ packages are included in the distribution repositories, the in
 
 Please see the [official RabbitMQ installation guide][2] for recommended installation strategies.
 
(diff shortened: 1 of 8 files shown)


apaskulin (approved):
In https://github.com/sensu/sensu-docs/pull/608 we kept the Erlang section, just modified it to be more general. If it's still appropriate to include here, I think it could be helpful for users.

+1 to copy to all versions

asachs01:
@apaskulin so the RMQ docs have the installation for Erlang in their docs. I'm more in favor of not including those in our docs (lest we have to deal w/ a package name change).

asachs01:
Also, by doing a `sudo apt-get install rabbitmq-server`, the requisite Erlang packages will be installed. I do think it would be a good idea to maybe add a note, but since the packages will all be added, I'm not sure of the utility of keeping the `Install Erlang` section.

Issue #887, opened 2018-11-07
Remove references to Redis as a transport


asachs01:
I propose we remove all references to Redis as a transport for Sensu:

1. Because it's not supported for production
2. Because of https://github.com/sensu/sensu-transport/issues/31
3. It's not an advised best practice
4. Having it as an option when it is obviously broken leads to a poor user experience and an influx of support cases where we have to say, "Sorry, but this isn't a best practice"

PR #888, opened 2018-11-07
Adding danger class

asachs01:
## Description
Adds a style for any note prefixed with `DANGER`. This is in relation to #887 . Rather than removing the content entirely, since this is a feature of the software, this will instead call attention to the fact that using RMQ as the transport is our recommended best practice. 

## Review Instructions
Add a note with `_DANGER: THIS IS A DANGER MESSAGE_` to content. This should show up like:

danger


apaskulin:
I think this use case is covered by the existing WARNING callout style. In general, DANGER callouts are reserved for issues of harm to people or property.

PR #890, opened 2018-11-08
Update RBAC docs for Sensu Go

### LINKED ISSUE #852: Replace organization/environment with namespaces
- Replace instances of "organization" and "environments" in configuration and definitions with "namespace"
- Include explanation of namespaces and best practices

- Defaults: default
- Standard sensuctl commands
- “Names of resources need to be unique within a namespace, but not across namespaces.”
- Ops team has checks that they maintain. Service team has the same. Ops wants to surface visibility, they can share read-only access to each other’s resources.
- Separate business units having access to Sensu within the same cluster
- Multitenancy

Resources:

- https://github.com/sensu/sensu-go/issues/2253 
- https://github.com/sensu/sensu-go/pull/2254 
- https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/

### LINKED ISSUE #873: Update Sensu Go RBAC model
https://github.com/sensu/sensu-go/issues/2262

To do:

- [ ] Update RBAC reference
- [ ] Update RBAC guide
- [ ] Search for other mentions of RBAC in the docs

apaskulin:
## Description

- Replaces organization and environment with namespace
- Adds new role and role binding types based on GoogleDoc
- Updates resource types based on https://github.com/sensu/sensu-go/blob/master/backend/apid/graphql/schema/rbac.graphql

To do:
- [x] Add resources to the glossary
- [x] Remove sensu default user

## Motivation and Context

Closes #852 
Closes #873

## Notes
https://github.com/sensu/sensu-go/blob/feature/rbac/types/rbac.go


content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
+| `checks` | [Check][6] resources within a namespace |
+| `entities` | [Entity][7] resources within a namespace |
+| `events` | [Event][8] resources within a namespace |
+| `handlers` | [Handler][9] resources within a namespace |
+| `hooks` | [Hook][10] resources within a namespace |
+| `mutators` | [Mutator][11] resources within a namespace |
+| `namespace` | [Namespaces][12] |
+| `roles` | [Roles][13] |
+| `silenced` | [Silencing][14] resources within a namespace |
+| `users` | [Users][15] |
```
apaskulin: Should there be a filters type included here? Is this the best source for a list of these types https://github.com/sensu/sensu-go/blob/master/backend/apid/graphql/schema/rbac.graphql ?

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
+Use the `admin` user to manage Sensu and create new users; use the `sensu` user to explore Sensu as a read-only user.
+User access credentials can be used to configure [sensuctl][2] or log in to the [dashboard][3].
 
-An organization is the top-level resource for RBAC. Each organization can
-contain one or multiple environments. Sensu ships with a `default` organization.
+| username | password    |  role       | access |
+| -------- | ----------- | ----------- | ------ |
+| `admin`  | `P@ssw0rd!` | `admin`     | Create, read, update, and delete access for all [resource types][4]. |
+| `sensu`  | `sensu`     | `read-only` | Read-only access for all [resource types][4]. |
+| `agent`  | n/a         | `agent`     | The `agent` user is used by the Sensu agent and doesn't have an accessible role or password. |
```
palourde: We made some changes around the default RBAC resources created, here's a summary with an exhaustive description:

| ClusterRole | ClusterRoleBinding | group | user | password | description |
| ----------- | ------------------ | ----- | ---- | ---------| ----------- |
| `cluster-admin` | `cluster-admin` | `cluster-admins` | `admin` | `P@ssw0rd!` | It gives access to perform any action on any When used in a ClusterRoleBinding, it gives full control over every resource in the cluster and in all namespaces. When used in a RoleBinding, it gives full control over every resource in the rolebinding's  namespace, including the namespace itself |
| `admin` | N/A | N/A | N/A | N/A | It is intended to be used within a namespace using a RoleBinding. It gives full access to most resources, including the ability to create Roles and RoleBindings within the namespace but does not allow write access to the namespace itself |
| `edit` |  N/A | N/A | N/A | N/A | It is intended to be used within a namespace using a RoleBinding. It allows read/write access to most objects in a namespace. It does not allow viewing or modifying roles or rolebindings. |
| `view` |  N/A | N/A | N/A | N/A | It is intended to be used within a namespace using a  RoleBinding. It allows read-only access to see most objects in a namespace. It does not allow viewing roles or rolebindings. |
| `system:agent` | `system:agent` | `system:agents` | `agent` | `P@ssw0rd!` | It is used by Sensu agents and should not be modified by the users. Modification to his ClusterRole can result in non-functional Sensu agents. |

For reference, these are basically the same as Kubernetes default resources: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
-## How does Role Based Access Control work?
-Sensu RBAC allows management and access of users and resources based on a heirarchy of *organizations*, *environments*, *roles*, *rules*, and *users*. Each `environment` belongs to only one `organization`, and each `resource` (user, check, asset, etc) belongs to only one `environment`. A Sensu installation can have multiple organizations, each with their own set of environments and resources belonging to them.
+Sensu role-based access control (RBAC) helps different teams and projects share a Sensu instance.
+RBAC allows management and access of users and resources based on a hierarchy of **namespaces**, **roles**, **users**, and **role bindings**.
 
-### Authentication
+- **Namespace**: Partitions resources within Sensu. Each Sensu resource (check, handler, etc.) belongs to a single namespace.
+- **Role**: Creates a set of permissions (read, delete, etc.) tied to resource types.
+- **User**: Represents a person or agent that interacts with Sensu.
+- **Role binding**: Assigns a role to a user.
```
palourde: We are missing `ClusterRole` & `ClusterRoleBinding`

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
 
-Sensu 2.0 offers local users management. Users can be managed with `sensuctl`.
+Sensu access controls apply both to [sensuctl][2] and the Sensu [dashboard][3].
 
-## Terminology
+## Default access
+Every [Sensu backend][1] includes a `default` namespace.
+All resources created without a specified namespace are created within the `default` namespace.
```
palourde: 👍

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
-description  | The name of the organization. Cannot contain special characters.
-required     | true 
-type         | String
-example      | {{< highlight shell >}}"name": "default"{{</highlight>}}
+## Namespaces
+Namespaces help teams use different resources (checks, handlers, etc.) within Sensu and impose their own controls on those resources.
+A Sensu instance can have multiple namespaces, each with their own set of managed resources.
+Resource names need to be unique within a namespace, but not across namespaces.
+
+### Default namespace
```
palourde, line 37: It seems duplicated with what we have at line 29!

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
+───────── ─────────────────── 
+default   Default namespace
+{{</highlight>}}
+
+### Creating a namespace
+
+You can use [sensuctl][2] to create a namespace:
+
+{{< highlight shell >}}
+sensuctl namespace create --interactive
```
palourde: We had a discussion internally around the interactive mode, and the outcome was that basically we will slowly move away from the interactive mode. Since this command is fairly simple, I'd recommend to simply do something like `sensuctl namespace create ops` instead. WDYT?

Also, I just realized there's no `description` field for namespaces. I'll ask around to see if people think we should add that back but I'm curious to have your opinion on that!

content/sensu-go/5.0/reference/rbac.md
palourde: The graphql stuff might not be the most up-to-date reference. We have a "registry" [here](https://github.com/sensu/sensu-go/blob/master/types/typemap.go#L11) but the challenge is that some of these types are embedded within others (e.g. a _Rule_ is part of a _Role_) and others are meant for internal use (e.g. _Claims_). The [APId routers](https://github.com/sensu/sensu-go/tree/master/backend/apid/routers) are probably a good place to get started too.

~Also, we introduced the notion of a `kind`, so we have to make a distinction between a `type`, a `kind` and a `resource`. Based on my understanding of Kubernetes, here's my definition for each of these terms but it's confusing and should be improved~

Update: We are finally giving up on the idea of introducing `kind`. (see https://github.com/sensu/sensu-go/issues/2327).

| Term | Description |
|---|---|
| object | Objects are persistent entities* in the Sensu system ([source](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#understanding-kubernetes-objects)) |
| resource | A representation of a system entity*, sent or retrieved as JSON via HTTP to the server. Users will specify _resources_ in their RBAC rules (e.g. entities, checks) | 
| type | The literal Go type. The term "type" should be used for distinguishing sub-categories within objects (e.g. An event object has a `check` and an `entity` type). | 
| ~kind~ | ~The name of a particular object schema (e.g. Check or Hooks kinds that have different attributes and properties). Users will need to specify a _kind_ when creating resources via `sensuctl create` for example.~ | 

*The term `entity` is really confusing here since we have Sensu entities, so we should probably find a better term here.

Anyway, I think we are missing the following resources:

- `cluster` (to manage the cluster)
- `filters`
- `extensions`
- `clusterroles`
- `clusterrolebindings`
- `rolebindings`

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
+|---|---|
+| `*` | All resources within Sensu. **The `*` type takes precedence over other rules within the same role.** If you wish to deny a certain type, you can't use the `*` type and must explicitly allow every type required |
+| `assets` | [Asset][5] resources within a namespace |
+| `checks` | [Check][6] resources within a namespace |
+| `entities` | [Entity][7] resources within a namespace |
+| `events` | [Event][8] resources within a namespace |
+| `handlers` | [Handler][9] resources within a namespace |
+| `hooks` | [Hook][10] resources within a namespace |
+| `mutators` | [Mutator][11] resources within a namespace |
+| `namespace` | [Namespaces][12] |
```
palourde: s/namespace/namespaces/

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
+### Assigning a resource to a namespace
+
+You can assign a resource to a namespace in the resource definition.
+Namespaces, users, and roles are part of Sensu RBAC and don't need to be assigned to a namespace.
+
+For example, to assign a check called `check-cpu` to the `ops` namespace, include the `namespace` attribute in the check definition:
+
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
```
palourde: ~Will soon be replaced by `kind`, see https://github.com/sensu/sensu-go/issues/2320~

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
+
+## Roles
+
+A role is a set of permissions controlling access to Sensu resources.
+Roles can be assigned to one or multiple users, and each user can be a member of one or multiple roles.
+Users inherit all of the permissions from each role they are
+in.
 
-#### Attributes
+### Default roles
```
palourde, line 325: See the default RBAC resources above

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,132 +1,298 @@
...
+    {
+      "verbs": ["create", "read", "update", "delete"],
+      "resources": ["*"],
+      "resourceNames": [""]
+    }
+  ]
+}
+{{</highlight>}}
+
+### Creating a cluster-wide role
```
palourde, line 382: It would actually a ClusterRole/ClusterRoleBinding. I realize it's really confusing at first, let me know if you would like more details about that!

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -202,20 +344,168 @@ example      | {{< highlight shell >}}
...
-| `events` | Manage event resources within a given organization & environment |
-| `handlers` | Manage handler resources within a given organization & environment |
-| `mutators` | Manage mutator resources within a given organization & environment |
-| `organizations` | Create and remove organizations |
-| `roles` | Create, remove roles and set rules within |
-| `users` | Manage user resources |
+## Users
+A user represents a person or an agent which interacts with Sensu.
+
+### Default users
```
palourde: See the default RBAC resources above

palourde (commented):
As promised, a bit more information about the new RBAC model!

content/sensu-go/5.0/reference/rbac.md
apaskulin: Good to know! I'll change this to `sensuctl namespace create ops` and remove the description field.

content/sensu-go/5.0/reference/rbac.md
apaskulin: Updated!

content/sensu-go/5.0/reference/rbac.md
apaskulin: Updated!

content/sensu-go/5.0/reference/rbac.md
apaskulin: Updated!

content/sensu-go/5.0/reference/rbac.md
apaskulin: Updated!

content/sensu-go/5.0/reference/rbac.md
apaskulin: I removed the defaults section and moved defaults content within the individual sections.

content/sensu-go/5.0/guides/create-read-only-user.md
```diff
@@ -11,49 +11,54 @@ menu:
...
+Since this user needs different permissions from those provided by the default `view` role, you'll need to create a role before creating the user and role binding.
 
-Using sensuctl, you can verify that the role and user have the correct permissions.
+1. Create an `event-reader` role with `get` permissions for `events` within the `default` namespace:
 {{< highlight shell >}}
-$ sensuctl role list-rules read-only
-  Type     Org.      Env.    Permissions
- ────── ───────── ───────── ─────────────
-  *      default    default     read
+sensuctl role create event-reader --get --namespace default --type events
```
palourde: The role command flags have changed a bit:

```
$ sensuctl role create -h
create a new Role with a single rule

Usage:	sensuctl role create [NAME] --verb=VERBS --resource=RESOURCES [--resource-name=RESOURCE_NAMES] [flags]

Flags:
  -h, --help                    help for create
  -r, --resource strings        resources that the rule applies to
  -n, --resource-name strings   optional resource names that the rule applies to
  -v, --verb strings            verbs that apply to the resources contained in the rule

Global Flags:
      --api-url string      host URL of Sensu installation
      --cache-dir string    path to directory containing cache & temporary files (default "/Users/splourde/Library/Caches/sensu/sensuctl")
      --config-dir string   path to directory containing configuration files (default "/Users/splourde/.config/sensu/sensuctl")
      --namespace string    namespace in which we perform actions (default "default")
```

The equivalent now would look like this: `sensuctl role create event-reader --verb get --verb list --resource events --namespace default` (you could also do `--verb get,list`)

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+Sensu role-based access control (RBAC) helps different teams and projects share a Sensu instance.
+RBAC allows management and access of users and resources based on **namespaces**, **roles**, **users**, and **bindings**.
 
-### Authentication
+- **Namespaces** partition resources within Sensu. Sensu checks, handlers, and other [namespaced resources][17] belong to a single namespace.
+- **Roles** create sets of permissions (get, delete, etc.) tied to resource types. **Cluster roles** apply permissions across namespaces and include access to [cluster-wide resources][18] like users and namespaces. 
+- **Users** represent a person or agent that interacts with Sensu. Users can belong to one or more **groups**.
+- **Role bindings** assign a role to a user, and **cluster role bindings** assign a cluster role to a user.
```
palourde: Two things that might be worth mentioning but not sure how and if it's relevant:

1. Bindings can assign roles to one or multiple users and/or one or multiple groups

2. A role binding can assign both roles and cluster roles to the subjects. In the latter case, what will happen is that the permissions defined in the cluster roles are basically restricted to the role binding's namespace. This is a useful pattern when you want to apply the same permissions across multiple namespaces; you simply create a cluster role and you assign it in the role bindings instead of creating the same role within all these namespaces.

Let me know if you have any questions!

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+Cluster-wide resources cannot be assigned to a namespace and can only be accessed by [cluster roles][21].
+
+| Type | Description |
+|---|---|
+| `cluster`   | Sensu clusters running multiple [Sensu backends][1] |
+| `clusterrolebindings`   | Cluster-wide role assigners  |
+| `clusterroles`   | Cluster-wide permission sets  |
+| `namespaces` | Resource partitions within a Sensu instance |
+| `rolebindings`   | Namespace-specific role assigners  |
+| `roles` | Namespace-specific permission sets |
```
palourde: rolebindings & roles are actually scoped to a namespace!

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+}
+{{< /highlight >}}
+
+### Creating a cluster-wide role
+
+You can use [sensuctl][2] to create a cluster role.
+For example, the following command creates a global event reader role that can read only events across all namespaces within Sensu.
+
+{{< highlight shell >}}
+sensuctl cluster-role create global-event-reader --read --type events
```
palourde: `sensuctl cluster-role create global-event-reader --verb get --verb list --resource events`

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+
+You can also create a role using a JSON role definition.
+
+{{< highlight shell >}}
+{
+  "name": "prod-admin",
+  "namespace": "production",
+  "rules": [
+    {
+      "verbs": ["create", "read", "update", "delete"],
```
palourde: `read` has been split into two verbs: `get` & `list`

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+sensuctl cluster-role list
+{{< /highlight >}}
+
+### Creating a role
+
+You can use [sensuctl][2] to create a role.
+For example, the following command creates an admin role restricted to the production namespace.
+
+{{< highlight shell >}}
+sensuctl role create prod-admin --create --get --update --delete --namespace production --type *
```
palourde: sensuctl role create prod-admin --verb '*' --resource  '*' --namespace production

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+{{< /highlight >}}
+
+You can also create a role using a JSON role definition.
+
+{{< highlight json >}}
+{
+  "name": "global-event-reader",
+  "rules": [
+    {
+      "verbs": ["get"],
```
palourde: We should probably add `"list"` to that list

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+
+### Managing roles
+
+You can use [sensuctl][2] to view, create, update, and delete roles.
+To use any of these commands with cluster roles, substitute the `cluster-role` command for the `role` command.
+
+To add permissions to a role:
+
+{{< highlight shell >}}
+sensuctl role add-rule [ROLE-NAME] [flags]
```
palourde: We have a ton of technical debt in sensuctl so we decided to drop the `add-rule` subcommand and other similar commands. Instead, Eric introduced a new command `sensuctl edit` which could be use to achieve the same result with more consistency. So instead of this subcommand, users could do `sensuctl edit roles [ROLE-NAME] [flags]` and edit the rules with their editor.

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+{{< /highlight >}}
+
+To remove permissions from a role:
+
+{{< highlight shell >}}
+sensuctl role remove-rule [ROLE-NAME] [flags]
```
palourde: See comment above ^

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+  }
+]{{< /highlight >}}
+
+#### Rule attributes
+A rule is an explicit statement which grants a particular permission to a resource.
 
-### User
+verbs  | 
+-------------|------ 
+description  | The permissions to be applied by the rule: `create`, `get`, `update`, or `delete`.
```
palourde: `list` is missing to this list

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -1,85 +1,407 @@
...
+{{< highlight shell >}}
+sensuctl user list
+{{< /highlight >}}
+
+### Creating a user
+You can use [sensuctl][2] to create a user.
+For example, the following command creates a user with the username `alice` and the password `P@ssw0rd!`.
+
+{{< highlight shell >}}
+sensuctl user create alice --password P@ssw0rd!
```
palourde: Probably safer to add quotes around the password because it contains special characters:

`sensuctl user create alice --password 'P@ssw0rd!'`

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -88,134 +410,227 @@ username     |
...
+To remove a user from all groups:
+
+{{< highlight shell >}}
+sensuctl user remove-groups [USERNAME]
+{{< /highlight >}}
+
+## Role bindings and cluster role bindings
+
+A **role binding** assigns a **role** to a user or set of users.
+A **cluster role binding** assigns a **cluster role** to a user or set of users.
```
palourde, line 487: See my comment at line 26 around assigning roles to cluster role bindings

content/sensu-go/5.0/guides/create-read-only-user.md
```diff
@@ -11,49 +11,54 @@ menu:
...
-$ sensuctl user create 'read-only-user'  --password 'password' --roles 'read-only'
+sensuctl user create alice --password 'password'
 {{< /highlight >}}
 
-You may also apply the read-only role to an existing user account. Users can
-have multiple roles spanning multiple organizations, each with their own set of
-permissions.
+2. Create a `read-only-user` role binding to assign the `view` role to the `alice` user:
+{{< highlight shell >}}
+sensuctl role-binding create read-only-user --role=view --user=alice
```
palourde: The `view` role is actually a Cluster Role so the flag should be `--cluster-role=view`.

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -88,134 +235,409 @@ username     |
...
+### Default roles
+
+Every [Sensu backend][1] includes:
+
+| Role name       | Type          | Description |
+| --------------- | ------------- | ----------- |
+| `cluster-admin` | `ClusterRole` | Full access to all [resource types][4] across namespaces, including access to [cluster-wide resource types][18].
+| `admin`         | `Role`        | Full access to all [resource types][4] within a namespace. |
+| `edit`          | `Role`        | Read and write access to most resources within a namespace with the exception of roles and role bindings.
+| `view`          | `Role`        | Read-only permission to most [resource types][4] within a namespace with the exception of roles and role bindings.  |
```
palourde: The `admin`, `edit` & `view` roles are all `ClusterRole`. The distinction here is that these 3 cluster roles are intended to be used via a Role Binding, so they act as a kind of template for each namespace.

content/sensu-go/5.0/reference/rbac.md
```diff
@@ -88,134 +235,409 @@ username     |
...
+{{< highlight shell >}}
+sensuctl cluster-role-binding list
+{{< /highlight >}}
+
+### Creating a role binding
+
+You can use [sensuctl][2] to see a create a role binding:
+
+{{< highlight shell >}}
+sensuctl role-binding create [NAME] --role=NAME [--user=username] [--group=groupname]
```
palourde, line 524: We should probably indicate that a role binding and either take a role or a cluster-role (e.g. `--cluster-role=NAME|--role=NAME`)

palourde (approved):
Just small details, otherwise LGTM!

PR #905, opened 2018-11-15
Fix example highlight for user

rgeniesse:
## Description
Add shell highlight tag for user example.

## Motivation and Context
Consistency with all other example rows.

## Review Instructions
Ensure proper syntax highlighting if you would like. Confirmed on my end.


apaskulin (approved):
:tada:

PR #927, opened 2018-11-26
Specify the RBAC requirements for --all-namespaces

palourde:
## Description
Specify the requirements for using the `--all-namespaces` flag in sensuctl.

## Motivation and Context
A user need a cluster role in conjunction with a cluster role binding for the specified resource (e.g. checks) to list the resource across all namespaces.


apaskulin (approved):
✨💖✨

PR #941, opened 2018-11-29
Adjust reference docs after metadata changes

echlebek:
We now support placing metadata outside of the spec of a resource.
Object metadata will still be usable in the spec by filters, but
will display in the wrapper.

If users put object metadata in both places, the metadata will be
merged, with a preference for the metadata keys in the wrapper in
case of a conflict.

Some lingering filter errata have been dealt with as well.

Signed-off-by: Eric Chlebek <eric@sensu.io>


apaskulin:
This is totally my fault, apologies for the duplicated work. Closing in favor of #943

Issue #580, opened 2018-07-16
Epic: Sensu Go API docs


apaskulin:
This is a tracking ticket for all planned docs within the API section. Notes below should be moved to individual tickets once created.

- [godoc](https://godoc.org/github.com/sensu/sensu-go/backend/apid)
- [source](https://github.com/sensu/sensu-go/tree/master/backend/apid/routers), ex: [checks](https://github.com/sensu/sensu-go/blob/master/backend/apid/routers/routers.go#L138-L145)
- https://github.com/sensu/sensu-go/pull/198

To do:

- [x] Create an API docs landing page
- [x] Research API doc generation tools for Go
  - https://goswagger.io/
  - https://swaggo.github.io/swaggo.io/
- [x] Document limit for data size returned
- [x] Document authentication process

Source:

- Endpoints: https://github.com/sensu/sensu-go/tree/master/backend/apid/routers
- Response codes: https://github.com/sensu/sensu-go/tree/master/backend/apid/actions
- Types: https://github.com/sensu/sensu-go/blob/master/api/core/v2/asset.pb.go
  - API types don't include the top-level `type` and `spec` scopes
  - /api/core/v2/namespaces/default/checks/my-check
- Routers: https://github.com/sensu/sensu-go/blob/master/backend/apid/apid.go

To write:
- [x] Overview page
- [x] Assets
   - [x] Get all
   - [x] Get
   - [x] Post
   - [x] Put
- [x] Checks
   - [x] Get all
   - [x] Get
   - [x] Post
   - [x] Put (needs example)
   - [x] Delete 
   - [x] Post id/execute Queue adhoc check request
   - [x] Put id/hooks/type
   - [x] Delete id/hooks/type/hook/hookid
- [x] Entities
   - [x] Get all
   - [x] Get
   - [x] Post
   - [x] Put
   - [x] Delete 
- [x] Filters
   - [x] Get all
   - [x] Get
   - [x] Post
   - [x] Put
   - [x] Delete 
- [x] Events
  - [x] Get all
  - [x] Get by entity /entity-name
  - [x] Get /entity-name/check-name
  - [x] Delete /entity-name/check-name
  - [x] Put /entity-name/check-name
  - [x] Post
- [x] Handlers
   - [x] Get all
   - [x] Get
   - [x] Post
   - [x] Put
   - [x] Delete 
- [x] Hooks
   - [x] Get all
   - [x] Get
   - [x] Post
   - [x] Put
   - [x] Delete 
[...]

PR #965, opened 2018-11-30
Fix Sensu Go PUT endpoints, add DELETE endpoints

apaskulin:
## Description

- Fixes PUT endpoints to require a resource name in the URL
- Adds DELETE endpoints

## Motivation and Context

Contributes to #580


nikkictl (approved):
👏

PR #974, opened 2018-12-04
Remove navigation button

### LINKED ISSUE #938: Remove Navigation button
Remove the Navigation button from the primary layout and replace it by linking to the navigation page from the header.

apaskulin:
## Description

This PR removes the navigation button from the top of the sidebar in favor of using the "Sensu Docs" title in the header as a link to the main page of the docs.

## Motivation and Context

Closes #938


asachs01:
I don't know if it's quite ready yet. The space where the navigation text was feels a bit...off?

Current:
image

Proposed:
image

What's weird to me is that the product text is smaller than the menu text. I would think it should be bigger (this isn't from this pr, obv), but with the change, it feels like that should be bigger.

@apaskulin just so I'm clear on the proposal, we're aiming for the area here:

image

To be the main place to click to take you back to the root of the site?

asachs01:
One other thing to do would maybe to be to remove the title of the product showing up above the nav, and instead, keep the "Navigation"  bit, but change that so that it is the product name, and takes you back to the product root?

apaskulin:
I just noticed that the ability to click on the product name at the top of the sidebar is broken in staging.

apaskulin:
Fixed and redeployed with fix for product name links working.

@asachs01 
> just so I'm clear on the proposal, we're aiming for the area here: To be the main place to click to take you back to the root of the site?

Yes.

> The space where the navigation text was feels a bit...off?

I think this looks fine as-is, but feel free to push CSS changes to this branch.

cwjohnston (approved):

PR #1004, opened 2018-12-10
Update token sub values to not use dashes.

### LINKED ISSUE #999: Use underscores in token substitution examples
It turns out that dashes don't (and can't) work in token substitution. We should change the examples.

rgeniesse:
## Description
Fixes example labels in tokens using `-`.

## Motivation and Context
Partially addresses #999 

## Review Instructions
Test out the config examples in Sensu 5.0 environment.


apaskulin (approved):
:tada:

PR #1009, opened 2018-12-12
Add remaining Sensu Go endpoints

apaskulin:
## Description

Adds remaining endpoints for the following Sensu Go APIs: Checks, cluster, and users.

## Motivation and Context

Closes #580

PR #1029, opened 2018-12-17
Correct behavior for DELETE users endpoint

apaskulin:
## Description

Changes description for users DELETE endpoint to disable instead of delete

## Motivation and Context

https://github.com/sensu/sensu-go/issues/2540


apaskulin:
Oops! This is covered by #1009

PR #1032, opened 2018-12-17
Clarify labels and tokens

### LINKED ISSUE #995: Clarify process for adding labels in the token reference
The token reference relies on entity labels for its examples, but doesn't indicate how to add labels to agents using the agent config file. Update the token reference to make this process more clear.

### LINKED ISSUE #999: Use underscores in token substitution examples

apaskulin:
## Description

This PR addresses some confusion around labels and tokens.

- Changes the reference docs for each type to specify character limits for label keys
- Adds a section about managing entity labels to the entity reference
- Updates the token reference to link to the new section in the entity reference
- Makes a few stylistic and formatting changes

## Motivation and Context

Closes #995 
Closes #999

### DIFF
--- a/content/sensu-go/5.0/reference/entities.md
+++ b/content/sensu-go/5.0/reference/entities.md
@@ -10,25 +10,112 @@ menu:
     parent: reference
 ---
 
+- [How do entities work?](#how-do-entities-work)
+- [Proxy entities](#proxy-entities)
+- [Managing entity labels](#managing-entity-labels)
 - [Entities specification](#entities-specification)
-	- [Entity Attributes](#entity-attributes)
-	- [System Attributes](#system-attributes)
-	- [Network Attributes](#network-attributes)
-	- [NetworkInterface Attributes](#networkinterface-attributes)
-	- [Deregistration Attributes](#deregistration-attributes)
+	- [Entity attributes](#entity-attributes)
+	- [System attributes](#system-attributes)
+	- [Network attributes](#network-attributes)
+	- [NetworkInterface attributes](#networkinterface-attributes)
+	- [Deregistration attributes](#deregistration-attributes)
 - [Examples](#examples)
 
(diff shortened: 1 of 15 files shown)


content/sensu-go/5.0/reference/entities.md
```diff
@@ -10,25 +10,112 @@ menu:
...
 
-## Proxy Entities
+An entity represents anything (ex: server, container, network switch) that needs to be monitored, including the full range of infrastructure, runtime and application types that compose a complete monitoring environment (from server hardware to serverless functions).
+We call these monitored parts of an infrastructure "entities".
+An entity not only provides context to event data (what/where the event is from) but an event's uniqueness is determined by the check name and the ID of the entity upon which the check ran.
```
echlebek: We recently transitioned from "ID" to "name" for entities, as of GA. We should avoid mentioning ID in the context of entities in the future.

echlebek (changes requested):
I think the changes from
```
us-west-2
```
to
```
us_west2
```
can be reverted, as label values can contain any valid UTF-8 string.

echlebek:
Looks good in general, the only thing we need to change is any mention of "ID". Before GA, we decided we would use only "name" and not use "ID" anywhere in the project.

content/sensu-go/5.0/reference/entities.md
apaskulin:
```suggestion
An entity not only provides context to event data (what/where the event is from) but an event's uniqueness is determined by the check name and the name of the entity upon which the check ran.
```

echlebek (approved):

apaskulin:
Thanks, @echlebek! I've updated ID to name, and reverted the change to us-west-x.

PR #1046, opened 2018-12-19
Users are not namespaced

palourde:
Signed-off-by: Simon Plourde <simon.plourde@gmail.com>

## Description
Correct the users API docs to reflect the fact that users are not namespaced.


apaskulin (approved):
⭐️

PR #1059, opened 2018-12-26
Securing Sensu Go requires modification of api-url parameter

cwjohnston:
## Description

Adds `api-url` parameter to the Securing Sensu guide

## Motivation and Context

Upgrading from beta to GA I ran into the problem described in sensu/sensu-go#2058. Specifying the https URL as the value of `api-url` solved my problem, and seems to be required for a secure configuration.

## Review Instructions

Note that I added `backend` to some of the other example yaml blocks. I think this helps beter differentiate backend config changes from agent changes. LMKWYT?


apaskulin (approved):
Approved based on pairing with @rgeniesse :shipit:

PR #1076, opened 2019-01-05
Errata fixes: new and improved sections, sensuctl quick reference

apaskulin:
## Description

- Break out the sensuctl quick reference into its own page
- Remove the "New and improved" sections of the reference docs in favor of the upgrade guide
- Correct the version in the example config


rgeniesse (approved):
🎉

PR #1081, opened 2019-01-08
Update Sensu query expression reference link

rgeniesse:
## Description
Link currently does not go to the sensu query expression reference.

## Motivation and Context
Make it so it does!

## Review Instructions
Ensure link is good.


apaskulin (approved):
⭐️

PR #1086, opened 2019-01-09
Fix example metric output_format configuration 

calebhailey:
For this example config to work, the `"type": "metric"` attribute would need to be set.

## Description
Added '"type": "metric"` to the example configuration snippet. 

## Motivation and Context
To help make it more obvious that `"type": "metric"` is required. 

## Review Instructions
It's a one-liner. :)


apaskulin (approved):
Great catch, thanks for fixing this! 🙌

calebhailey:
@apaskulin thanks for finishing this PR & merging! :P 🎉

Issue #865, opened 2018-11-02
Audit SE integrations for those that support proxy configurations


asachs01:
## This is an issue with:

- [ ] Bug (site functionality or styling)
- [ ] Errata (fix needed for doc content)
- [ ] New content (guide wanted)
- [x] Update (Add missing or refresh existing content)

## Expected Behavior
Describe integrations that allow for proxy attribute configuration

## Current Behavior
Some integrations have this capability, but it isn't called out

## Possible Solution
Audit the docs/integrations to determine which ones support proxy attributes.

asachs01:
## Integrations currently supporting explicitly defined proxy configurations
- [x] : Chef

## Integrations supporting proxy configuration via `http_proxy` environment variable
- [x] : Influxdb
- [x] : Jira 
- [x] : Librato
- [x] : Opsgenie
- [x] : Puppet
- [x] : Rollbar
- [x] : Servicenow

## Integrations currently NOT supporting proxy configurations
- [ ] : Datadog
- [ ] : Email
- [ ] : Graphite
- [ ] : Graylog
- [ ] : Hipchat
- [ ] : IRC
- [ ] : Opentsdb
- [ ] : Pagerduty
- [ ] : Slack
- [ ] : SNMP
- [ ] : Timescaledb
- [ ] : Victorops
- [ ] : Wavefront

## Integrations in question re: proxy configurations
- [ ] : EC2 - The integration doesn't explicitly spell out using a proxy. If I'm in AWS, do I even need to use one? What about if I'm using an on-prem Sensu cluster with the EC2 integration?

asachs01:
So the larger question here is how do we document these integrations, knowing that a handful only support using a http proxy via an environment variable, and one has its own proxy configuration? I'm not certain that documenting the workaround (env var) would serve our customers, or us well, knowing that if we make the decision to change how the http proxy configuration works for integrations, we'd have to then go and update the docs yet again. Personally, I'd rather update the docs once wrt the integrations, rather than essentially doubling the work. 

Though on the other hand, if this option is available, then perhaps it should be documented. @apaskulin do you have feels on this?

hillaryfraley:
Pending EOL: won't fix

PR #1090, opened 2019-01-10
Document http_proxy parameter for integrations which support it

cwjohnston:
## Description

Add documentation of the `http_proxy` attribute for Enterprise integrations which support it.

## Motivation and Context

Any Enterprise integration making use of our proprietary HTTPClient library is expected to honor the `http_proxy` attribute. Documentation for each integration using this library should indicate support for this attribute.

Currently the applicable integrations are:

- [x] ServiceNow
- [x] Puppet
- [x] Rollbar
- [x] Jira
- [x] OpsGenie
- [x] VictorOps (from 3.0 on)

Relates to https://github.com/sensu/sensu-docs/issues/865

### DIFF
--- a/content/sensu-enterprise/2.6/integrations/jira.md
+++ b/content/sensu-enterprise/2.6/integrations/jira.md
@@ -97,6 +97,13 @@ required     | false
 type         | String
 example      | {{< highlight shell >}}"root_url": "https://services.example.com/proxy/jira"{{< /highlight >}}
 
+http_proxy   | |
+-------------|------
+description  | The URL of a proxy to be used for HTTP requests.
+required     | false
+type         | String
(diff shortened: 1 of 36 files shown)


apaskulin (approved):
✅ Looks good to me! Seems like this PR updates Jira v2.6, Puppet v2.6, Rollbar v2.6, Servicenow v2.6, and OpsGenie v3.0, so we may need to copy to other versions before merging.

cwjohnston:
I have added the attribute to VictorOps and copied the change to all applicable versions of the other integrations, so I'm removing the WIP label.

Please note that OpsGenie docs prior to 3.0 have been replaced with notices indicating that Sensu Enterprise 3.0 or later is required for operation, so we don't have a need to add the http_proxy attribute to those versions.

apaskulin (approved):

PR #1093, opened 2019-01-10
Update `http_proxy` attributes in InfluxDB, Librato, and VictorOps integrations

apaskulin:
## Description

- Adds http_proxy attribute to influx and librato integrations
- Removes http_proxy from victorops integration

## Motivation and Context

Follow up to #1090


cwjohnston (approved):
Looks good, thanks! 👍

PR #1111, opened 2019-01-16
Update influx-db-metric-handler.md

mburgess00:
Updated the environment variables sent to the influxdb handler


apaskulin (approved):
This is a great catch, thanks @mburgess00! 💚 I've added a commit to make this change everywhere we mention these variables.

PR #1108, opened 2019-01-15
Add that agents are namespaced

rgeniesse:
## Description
Closes #1077

## Motivation and Context
Inform users that sensu agents can only belong to one namespace.

PR #1112, opened 2019-01-16
Agents are namespaced copy to all versions

rgeniesse:
## Description
- Add to all versions from #1108 
- Have `entities` listed above Namespace heading.

## Motivation and Context
Trigger happy.


apaskulin (approved):
💯

PR #1116, opened 2019-01-17
Update proxy request PR

apaskulin:
## Description

- Changes example in second section of the guide to use splay attributes
- Tries to provide more clarity in check reference for proxy request attributes


rgeniesse (approved):
LGTM. I'll check it out one more time in proxy-requests-2 locally before merging/after approval.

PR #1123, opened 2019-01-18
Update client-to-server transport documentation

gattie:
## Description

The example yaml file under the *Client-to-server transport security with HTTPS* section in the clustering guide says to use the client.pem and client-key.pem files. However, this fails to work and are not the correct certs and keys to use in this location. Instead, you should use the specific backend cert and keys you generated in the previous step.

## Motivation and Context

This change is required because currently the provided solution in the documentation leads users down a path that will fail. The change provided in this PR should direct users to utilize the correct certs and keys.

### DIFF
--- a/content/sensu-go/5.0/guides/clustering.md
+++ b/content/sensu-go/5.0/guides/clustering.md
@@ -256,13 +256,33 @@ client.pem
 
 ### Client-to-server transport security with HTTPS {#client-to-server-with-https}
 
-Below are example configuration snippets from `/etc/sensu/backend.yml` that will be the same for `backend-1`, `backend-2` and `backend-3`.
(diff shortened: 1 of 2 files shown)


rgeniesse:
Thanks for the PR @gattie! I have a vagrant setup I am going to run this through to confirm/update my tests, but based on the Slack conversation it sounds like you and other users have ran into the same issue.

gattie:
@rgeniesse your updates look good from my perspective. I think that addresses the issue of which certs and keys to use as well as providing an example config for each backend in the cluster 👍

rgeniesse:
Thanks @gattie. I've added them as agents connected to the cluster should have each backend in its configuration.

I'll copy to 5.1.

rgeniesse (approved):
I'll merge once PR CI comes back. For whatever reason, branch isn't working. Might be some new travis.com integration and forked repo weirdness.

PR #1124, opened 2019-01-22
Add missing subscriptions payload parameter

calebhailey:
Add missing `subscriptions` payload parameter to the `POST /checks/:check/execute` API docs.

## Description

The `POST /checks/:check/execute` API docs are missing the supported "subscriptions" payload parameter; see: https://github.com/sensu/sensu-go/blob/master/api/core/v2/adhoc.pb.go#L26-L28 

## Motivation and Context

Requesting adhoc executions of checks (against arbitrary subscriptions) is needed for various Sensu workflows including "automated remediation" and "self-healing" workflows. The most common use case is to target an entity subscription (i.e. a service check fails on one particular node; a remediation handler requests an action to be executed against that same node, where the "action" is a check execution configured with `"publish": false`, and the command is some remediation action such as starting or restarting a service).  

## Review Instructions

Test my changes and make sure they work? 😄

### DIFF
--- a/content/sensu-go/5.0/api/checks.md
+++ b/content/sensu-go/5.0/api/checks.md
@@ -320,6 +320,8 @@ HTTP/1.1 202 Accepted
 {"issued":1543861798}
 {{< /highlight >}}
 
+_PRO TIP: Include the `subscriptions` attribute with the request body to override the subscriptions configured in the check definition. This gives you the flexibility to execute a check on any Sensu entity or group of entities on demand._
+
 #### API Specification {#checkscheckexecute-post-specification}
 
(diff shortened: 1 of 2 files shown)


calebhailey:
This is a fix for what I thought was an omission in the actual API; see: https://github.com/sensu/sensu-go/issues/2624

apaskulin (commented):
Great catch, thanks @calebhailey! I think this is super valuable information for our users. I added a few notes to make the override behavior clear without being too specific. Let me know what you think.

rgeniesse (approved):
New parameters work for me as expected. 👍

PR #1134, opened 2019-01-24
Update silenced.md

rocdove:
Adds a regular match description for check and subscription


asachs01:
Hi @rocdove , 

Thanks for submitting this. Can you give us a bit more context around this change you're proposing?

apaskulin (approved):
Thanks, @rocdove! This is a great addition to these docs. 🌟 

I've added commits to update the check regex to match the code here https://github.com/sensu/sensu/blob/master/lib/sensu/api/routes/silenced.rb#L49-L50, update the formatting, and copy to all versions.

PR #1138, opened 2019-01-25
Correct client install next page link

rgeniesse:
## Description
Corrects a client install next page link that points to itself.

## Motivation and Context
Ensure instructions are correct.

## Review Instructions
Run locally and make sure link works. Tested locally myself. All other versions are correct.


apaskulin (approved):
Great catch! +1 to copy to all versions

PR #1221, opened 2019-02-20
Add note that /health does not require authentication and update command

### LINKED ISSUE #1202: Health API doesn't require authorization
Page: https://docs.sensu.io/sensu-go/5.2/api/health/

The health API doesn't require authorization, so remove this from the docs and possibly make a note on the overview page

rgeniesse:
## Description
Closes #1202

## Motivation and Context
Correct command that currently mentions use of a token. Also gives users a straight forward way to monitor the health of a backend without require auth (like for AWS ELBs).

## Review Instructions
- [ ] Copy to all versions

### DIFF
--- a/content/sensu-go/5.0/api/overview.md
+++ b/content/sensu-go/5.0/api/overview.md
@@ -39,11 +39,9 @@ Beta APIs, while more stable than alpha versions, offer similarly short-lived li
 
 ### Access control
 
-The Sensu API requires authentication using a JWT access token.
+With the exception of the [health API][5], the Sensu API requires authentication using a JWT access token.
 Sensuctl provides an easy way to generate access tokens for short-lived use with the Sensu API.
 The user credentials that you use to log in to senscutl determine your permissions to get, list, create, update, and delete resources using the Sensu API.
-
-_NOTE: The [`/health`][5] API endpoint does not require authorization._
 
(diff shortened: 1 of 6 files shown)


content/sensu-go/5.0/api/overview.md
```diff
@@ -43,6 +43,8 @@ The Sensu API requires authentication using a JWT access token.
 Sensuctl provides an easy way to generate access tokens for short-lived use with the Sensu API.
 The user credentials that you use to log in to senscutl determine your permissions to get, list, create, update, and delete resources using the Sensu API.
 
+_NOTE: The [`/health`][5] API endpoint does not require authorization._
```
apaskulin: I remember referring to this as a "note", so apologies for that. However, I don't think this information really fits in with our definition for NOTE callouts. I'd recommend integrating this into the preceding paragraph. Perhaps something like: `With the exception of the [health API][5], the Sensu API requires...`


Also, just noticed `senscutl` on line 44 (_facepalm_). If you're able to fix that as well, I'd be forever grateful 🙏

apaskulin (approved):
+1 to copy to all versions

content/sensu-go/5.0/api/overview.md
rgeniesse: Added the typo fix.

PR #1225, opened 2019-02-20
Document built-in deregistration extension

apaskulin:
## Description

- Clarifies install process for deregistration handler in v1.6 and earlier
- Calls out built-in deregistration extension in v1.7

To do:
- [x] Copy to v1.5 and earlier
- [x] Add extensions config

### DIFF
--- a/content/sensu-core/1.7/reference/clients.md
+++ b/content/sensu-core/1.7/reference/clients.md
@@ -112,7 +112,8 @@ registration events are logged in the [Sensu server][5] log._
 
 Similarly to registration events, Sensu can create and process a deregistration event when the Sensu client process stops.
 You can use deregistration events to trigger a handler that updates external CMDBs or performs an action to update ephemeral infrastructures.
-To enable deregistration events, set the `deregister` flag to `true` and specify the event handler using the [`deregistration` definition scope][48].
+By default, registration events trigger the built-in deregistration extension.
+To enable deregistration events, set the client `deregister` flag to `true` and specify the event handler using the client [`deregistration` definition scope][48].
 
 #### Proxy clients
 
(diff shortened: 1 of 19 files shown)


content/sensu-core/1.7/reference/clients.md
```diff
@@ -108,6 +108,12 @@ _WARNING: registration events are not stored in the event registry (in the Sensu
 [data store][8]), so they are not accessible via the Sensu API; however, all
 registration events are logged in the [Sensu server][5] log._
 
+#### Deregistration events
+
+Similarly to registration events, Sensu can create and process a deregistration event when the Sensu client process stops.
```
cwjohnston, line 113:
```suggestion
Similar to registration events, Sensu can create and process a deregistration event when the Sensu client process stops.
```

content/sensu-core/1.7/reference/clients.md
```diff
@@ -108,6 +108,12 @@ _WARNING: registration events are not stored in the event registry (in the Sensu
...
+You can use deregistration events to trigger a handler that updates external CMDBs or performs an action to update ephemeral infrastructures.
+To enable deregistration events, set the `deregister` flag to `true` and specify the event handler using the [`deregistration` definition scope][48].
```
cwjohnston: This line probably needs words to clarify that these settings are configured on the client side.

cwjohnston (changes requested):

content/sensu-core/1.7/reference/clients.md
apaskulin: "Similarly" is correct here because it's referring to "can create and process", not "Sensu" or "deregistration events".

cwjohnston (approved):

PR #1226, opened 2019-02-20
remove husky and precommit script

### LINKED ISSUE #881: Proposal: Remove commit hooks
This project currently uses commit hooks, meaning that the build is tested every time a commit is made on any branch. This is a nice way to check that a commit won't break the build, but I'd like to propose that we remove commit hooks for the following reasons:

1. Commit lag: In the last few months, the build time has increased by a few seconds, but it's just enough to make my workflow feel laggy.
2. User confusion: Even though we call out commit hooks in our contributing guide, they can still trip up new users who don't expect to see Hugo output when running `git commit`.
3. Redundancy with existing continuous integration pipeline

cwjohnston:
## Description

Remove husky dependency and `precommit` script from package.json

## Motivation and Context

Closes #881

Once this change is applied, it may be necessary for folks to run `node ./node_modules/husky/bin/uninstall.js` to remove existing git hooks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   "devDependencies": {
     "conzole": "^0.2.0",
     "grunt": "^1.0.1",
-    "husky": "^0.14.3",
     "js-yaml": "^3.9.1",
     "string": "^3.3.3",
     "toml": "^2.3.2",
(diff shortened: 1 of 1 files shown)


apaskulin (approved):
Ran this branch locally and the site builds successfully.

PR #1229, opened 2019-02-21
Copy to clipboard button

### LINKED ISSUE #832: Implement copy-to-clipboard button
## This is an issue with:

- [x] Feature request

#Comments 
It would save users a bit of time if we had a copy to clipboard button on our code samples. This should be easy enough to implement using something like https://themes.gohugo.io/theme/bulma/posts/clipboard/ .

asachs01:
## Description
Copying code bits, but easier

Also cleans up what we worked on in 832. But we have a button now.

## Motivation and Context
Closes #832


static/stylesheets/docs-overrides.css
```diff
@@ -1888,3 +1888,21 @@ img[alt=ssl_example] {
...
+  background-position: 50% 50%;
+  background-size: 16px 16px;
+  background-repeat: no-repeat;
+  width: 27px;
+  height: 1.45rem;
+  top: -1px;
+  display: inline-block;
+  vertical-align: middle;
+  position: relative;
+  color: #5e5e5e;
```
apaskulin, line 1901: Not sure why, but it doesn't seem to be respecting this.

apaskulin (approved):
Awesome work! The color looks great. I wasn't able to fix the extra line added for longer commands, but I don't think that needs to stop us from merging this.

PR #1244, opened 2019-02-26
Remove OSS designation from getting started docs, clarify open core

cwjohnston:
## Description

Removes the `OSS` designation from getting started docs

## Motivation and Context

The `OSS` designation creates the impression that there's some distinction between Sensu Go open source vs paid, when the enterprise features are in fact baked in to the freely downloadable artifacts we offer.


apaskulin (approved):
🙌

cwjohnston:
I think there are a few more places that need updates. Marking as WIP for now.

PR #1301, opened 2019-03-12
Call out dashboard namespace switcher

apaskulin:
## Description

- Adds a section to the dashboard overview calling out the namespace behavior and switcher
- Sentence case corrections

Screen Shot 2019-03-12 at 12 43 31 PM

## Review Instructions

- Image quality
- Better wording than "switcher"?


rgeniesse (commented):
Works locally for me! And the images don't look pixely. As for the "switcher", only other word I could come up with is navigator. The word switched is used with this Kubernetes issue that came up to make it easy to switch namespaces:

https://github.com/kubernetes/kubernetes/issues/40968

I see the word context as well. There's always dropdown, but that doesn't feel right. Dropdown describes the type of menu. While that dropdown does have namespaces in it, there's something more here to describe being "in" the namespace.

PR #1304, opened 2019-03-13
update agent and backend service verfication

rgeniesse:
## Description
The current command doesn't verify that the service is running, rather that the binary or package has been installed.

## Motivation and Context
As the previous command starts the service, the next command should be verifying it rather than confirming installation of the package or presence of the the binary.


apaskulin (approved):
🙌

PR #1310, opened 2019-03-14
Add Bonsai to plugin install instructions

### LINKED ISSUE #1286: Change ordering of sections in Assets reference
In the [Assets reference](https://docs.sensu.io/sensu-go/5.2/reference/assets/), the section "Sharing on asset on Bonsai" falls right in the middle of the Assets specification sections.  I believe it should be moved.
## This is an issue with:

## Expected Behavior

For readability,  the order for the [Assets reference](https://docs.sensu.io/sensu-go/5.2/reference/assets/) it should **not** document Bonsai right in the middle of Assets specification.
## Current Behavior

Bonsai is referenced in the middle of defining Asset specifications and seems disjointed.
## Possible Solution

If wanting to leave Bonsai in the Assets reference, at least move it down below the rest of the actual Assets documentation.

1.
2.
3.

## Context

It seems out of place and is confusing when working serially through the Assets reference documentation.
* Browser(s) and version(s): N/A
* OS: N/A

apaskulin:
## Description

- Adds Bonsai to the plugin install page
- Updates asset copy
- Copy asset content to all versions
- Moves Bonsai section to the end of the asset reference (closes #1286 )


content/sensu-go/5.0/reference/assets.md
```diff
@@ -11,25 +11,43 @@ menu:
...
 
 ## How do assets work?
-Assets can be executed by the backend (for handler and mutator assets), or
+Assets can be executed by the backend (for handler, filter, and mutator assets), or
 by the agent (for check assets). At runtime, the entity sequentially fetches
 assets and stores them in its local cache. Asset dependencies are then
 injected into the `PATH` so they are available when the command is executed.
-Subsequent check, handler, or mutator executions look for the asset in local
+Subsequent check, handler, filter, or mutator executions look for the asset in local
 cache and ensure the contents match the checksum. An entity's local cache can
```
rgeniesse: While not part of this change, this may need an update. It talks about `An entity's local cache`. Do we consider a sensu backend an entity? It can be one _if_ you have an agent. I bring this up as a sensu backend can have assets and if they aren't considered an entity, we may want to say something like `An agent or backend's local cache` instead.

I don't see the `cache-dir` flag in the backend reference. So are all assets downloaded by an entity when they are required? If so, that would mean assets for use on the backend are dependent on an agent to do that work and that should definitely be called out somewhere if it isn't already.

rgeniesse (approved):
I left a comment, but I don't think that should stop this PR, as this is for the Bonsai launch. If either cases are true (or maybe neither, :-)), probably should be a separate PR.

Great work. 💯

content/sensu-go/5.0/reference/assets.md
apaskulin: Great catch! I've updated the wording here and opened a PR to add `cache-dir` to the backend reference.

PR #1312, opened 2019-03-15
Adds Algolia search capabilities

### LINKED ISSUE #745: Site search field is slow to respond
Per @apaskulin:

> Since switching to full-text search, I've noticed that the site search has become pretty slow. It takes around 3 seconds for the search bar to become active once opened. This is a common issue with static search for larger sites: the more content you have, the larger the search index you have to load in-browser.

## This is an issue with:

- [x] Bug (site functionality or styling)

## Expected Behavior

The search field should appear, grab input focus, and **be responsive to user input** in a reasonable amount of time.

## Current Behavior

It can take upwards of a few seconds for the search field to show the flashing cursor and accept user input. This delay can be confusing to the user, who may give up and assume that site search is broken.

## Possible Solution

Per @apaskulin:

> Some sites (like https://gohugo.io/), get around this issue using a free tool called Algolia DocSearch https://community.algolia.com/docsearch/. I had hoped to put together a proof of concept for Algolia with the docs site, so I went ahead and requested a trial, but I've run up against my limited knowledge of Javascript. I don't think I'll be able to complete a proof of concept, but I have all the credentials if someone else wants to give it a try! There may be some difficulties getting their setup to work with our multi-product configuration, but it could be a good way to solve a lot of these issues.

1. Load https://docs.sensu.io/sensu-core/1.5/.
2. Click the search icon (:mag:).
3. Observe that it takes a few seconds for the cursor to appear in the search field.

asachs01:
## Description
Fixes #745 . The intent here is to get rid of the lunr search in favor of Algolia.


asachs01:
Efffffffff
```
Uncaught (in promise) 
AlgoliaSearchError
debugData: [{…}]
message: "Invalid Application-ID or API key"
name: "AlgoliaSearchError"
statusCode: 403
stack: "AlgoliaSearchError: Invalid Application-ID or API key↵    at success (https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js:2:81406)"
__proto__: Error
```

asachs01:
Ok, so some additional information as a result of this: the original apikey that was provided via our convo w/ Algolia was exposed. However, as that API key is defunct, search isn't working, and this leaves us with an opportunity to decide how we're going to handle secrets as part of the search implementation. We can do so via the following ways, as I understand it:

1.) Add it to footer_js and use dotenv
2.) Use the heroku add-on
3.) Implement it in a hugo partial and use site params

Options 1 & 3 seem the most labor intensive to me. Implementing the Heroku add-on seems the least labor intensive, and probably a natural extension given we're using Heroku.

cwjohnston:
> Options 1 & 3 seem the most labor intensive to me. Implementing the Heroku add-on seems the least labor intensive, and probably a natural extension given we're using Heroku.

In reviewing the documentation for the heroku add-on, I see that it generates API keys which need to be used in your app (or static site in our case) -- the add-on doesn't perform any magic tricks to insert API keys for us.

As a result, I am under the impression that regardless of the route we choose, the read-only API key will be visible in the site source code served to browsers. Is that right?

asachs01:
Having reverted from instantsearch back to docsearch (sorry, lots of internal convo that is left out here), things are working and we're going to go ahead and just push the key since it gets rendered client side anyways, regardless if we use env vars. We're just waiting on the crawler to crawl the site so search works.

cwjohnston (approved):

PR #1346, opened 2019-04-02
Correct RabbitMQ configuration example in SSL guide

rgeniesse:
## Description
Current RabbitMQ JSON example is missing the `rabbitmq` level.

## Motivation and Context
Users following this example will not have a working configuration and will have to know to open the RabbitMQ reference, asking in Community or open a support request.

## Review Instructions
I confirmed this works in my lab environment.


apaskulin (approved):
💚

PR #1349, opened 2019-04-03
Remove agent socket TTL example

### LINKED ISSUE #1348: Docs Feedback
Link to page: 

https://docs.sensu.io/sensu-go/5.4/reference/agent/#creating-a-dead-man-s-switch

I'm not sure if it is a bug in the documentation or a bug in sensu go, but it definitely doesn't work.

If it is a bug in the sensu go agent, I've filed a bug here:

https://github.com/sensu/sensu-go/issues/2858

If you could help move it along that would be great. If it is a bug in the documentation because the feature was withdrawn, this is a big impediment to migrating to sensu go as it is quite an important feature.

apaskulin:
## Description

The examples shown in https://docs.sensu.io/sensu-go/5.4/reference/agent/#creating-a-dead-man-s-switch do not work because the agent socket ignores all attributes that aren't listed in the [socket event specification](https://docs.sensu.io/sensu-go/5.4/reference/agent/#socket-event-specification). This PR removes the dead man's switch section from all versions of the agent reference. This content can be re-added following the resolution of https://github.com/sensu/sensu-go/issues/2858.

## Motivation and Context

Closes #1348

### DIFF
--- a/content/sensu-go/5.0/reference/agent.md
+++ b/content/sensu-go/5.0/reference/agent.md
@@ -276,32 +276,6 @@ default      | `1`
 type         | Integer
 example      | {{< highlight shell >}}"interval": 60{{< /highlight >}}
 
-### Creating a "dead man's switch"
-
-The Sensu agent socket in combination with check TTLs can be used to create what's commonly referred to as a "dead man's switch".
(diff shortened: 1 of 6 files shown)


rgeniesse (approved):
🎉

PR #1357, opened 2019-04-10
Correct config file examples for string slice flags

### LINKED ISSUE #1347: Redact string missing space between values
## This is an issue with:

## Expected Behavior
Example values for attributes are valid.

## Current Behavior
The example redact value in the agent reference guide is not value. Sensu Go will not properly parse the values to be redacted. 

https://docs.sensu.io/sensu-go/5.3/reference/agent/#security-configuration-flags

This:

```
redact: "secure-key,secure-password"
```
Should be:

```
redact: "secure-key, secure-password"
```

## Possible Solution
- [ ] Fix the issue described above
- [ ] Find other occurrences of this problem as other attributes may take comma delimited lists of strings.

apaskulin:
## Description

This PR corrects the docs for agent and backend config attributes of type string slice.

String slice attributes in backend config:
- etcd-advertise-client-urls
- etcd-listen-client-urls
- etcd-listen-peer-urls
- etcd-initial-advertise-peer-urls

String slice attributes in agent config:
- backend-url
- redact
- statsd-event-handlers
- subscriptions

To do:

- [x] Decide on a type with which to label these attributes. Options: String slice, list, array of strings
- [ ] Copy to all versions

## Motivation and Context

Closes #1347


content/sensu-go/5.5/reference/agent.md
```diff
@@ -572,14 +572,19 @@ Flags:
 | backend-url |      |
 --------------|------
 description   | ws or wss URL of the Sensu backend server. To specify multiple backends using `sensu-agent start`, use this flag multiple times.
-type          | String
+type          |
```
apaskulin: What type should we list here? String slice, list, array of strings?

palourde: I think we should avoid using `slice`. While it's not exclusive to Go, it's kind of specific so I wouldn't expect everyone to know this term.

I like `list` because it's high-level and it doesn't imply much; `StringSlice` flags can be used in many ways so it kind of reflect that, e.g.
- `--subscriptions foo`
- `--subscriptions foo,bar`
- `--subscriptions foo --subscriptions bar`

rgeniesse (approved):
LGTM, did some simple testing. 👍

palourde (approved):

PR #1355, opened 2019-04-10
Add sensu logging to /var/log/sensu/ on systemd based systems.

rgeniesse:
## Description
Closes #1292 

## Motivation and Context
Give users the option to have log files written to `/var/log/sensu/servicename.log` when `systemd` is the service manager. While `journalctl` does give you access to what you need, it's a nice option to have a more traditional logging system in place.

PR #1381, opened 2019-04-19
Updates frontmatter so content is rendered

asachs01:
## Description

#1355 added content for sending Sensu Go journald logging data to syslog, but the content is rendered. I've updated the frontmatter so this happens.

## Motivation and Context

We have a customer who needs to see the content.


apaskulin:
This page is discoverable via search and through the troubleshooting guide. We can revisit this if we hear of users not being able to find this content.

PR #1401, opened 2019-04-28
Correcting link in checks.md

agm650:
## Description
Correcting a link to target the "monitor external ressource" instead of "install check executables with assets"

## Motivation and Context
Ensuring the documentation is coherent.


apaskulin (approved):
Great catch, @agm650. Thanks for fixing this! I added a commit to copy your change to all versions of the docs. :shipit:

PR #1412, opened 2019-04-30
Correct license requirement

apaskulin:
## Description

Clarifies that an enterprise license is required to delete entities using the web UI.

## Motivation and Context

Tested using Sensu Go 5.6 on CentOS7


jamesdphillips (approved):
🎉

PR #1416, opened 2019-05-01
Add clickable anchors to heading elements

### LINKED ISSUE #604: Anchor link selector
Anchor links on headings let users link to specific content within the page. Currently, scrolling to a heading changes the URL, but the general best practice is to provide a button to select the heading and change the URL.

To do:
Make headings selectable (either using a button or by clicking on the heading itself) so that users can confidently share a link to a specific heading.

Design references:

- https://docs.microsoft.com/en-us/dotnet/visual-basic/getting-started/index#see-also
- https://gohugo.io/templates/introduction/#basic-syntax

Notes:
- https://github.com/gohugoio/hugoDocs/blob/master/themes/gohugoioTheme/assets/js/anchorforid.js

rnevius:
## Description
Appends anchor elements to headings with id attributes.
Ignores headings that are within a "platform" context.

## Motivation and Context
Improves the UI for linking to a specific part of a page.

Resolves #604

### DIFF
--- a/layouts/partials/footer_js.html
+++ b/layouts/partials/footer_js.html
@@ -45,7 +45,6 @@
         var a = document.createElement("a");
         a.setAttribute("class", "header-link");
         a.setAttribute("href", "#" + headers[i].id);
-        a.setAttribute("title", "Permanent link")
         a.innerHTML = '<svg height="22px" viewBox="0 0 24 24" width="22px" xmlns="http://www.w3.org/2000/svg"><path d="M0 0h24v24H0z" fill="none"/><path d="M3.9 12c0-1.71 1.39-3.1 3.1-3.1h4V7H7c-2.76 0-5 2.24-5 5s2.24 5 5 5h4v-1.9H7c-1.71 0-3.1-1.39-3.1-3.1zM8 13h8v-2H8v2zm9-6h-4v1.9h4c1.71 0 3.1 1.39 3.1 3.1s-1.39 3.1-3.1 3.1h-4V17h4c2.76 0 5-2.24 5-5s-2.24-5-5-5z"/></svg>';
         headers[i].appendChild(a);
       }
--- a/layouts/partials/footer_js.html
+++ b/layouts/partials/footer_js.html
@@ -1,0 +1,100 @@
+    <script>
+    {{ with .Scratch.Get "repo_id" }}
+      var base_url = '{{ (printf "%s" $.Site.BaseURL) | safeJS }}';
+      var repo_id  = '{{ . | safeJS }}';
+    {{ else }}
+      var base_url = '';
+      var repo_id  = '';
+    {{ end }}
+    </script>
+
+    <script src="{{ "/javascripts/application.js" | absURL }}"></script>
+    {{ range .Site.Params.custom_js }}
+    <script src="{{ . | absURL }}"></script>
+    {{ end }}
+
+    <script>
+      {{ "/* Add headers to scrollspy */" | safeJS }}
+      var headers   = document.getElementsByTagName("h2");
+      var scrollspy = document.getElementById('scrollspy');
+
+      if(scrollspy) {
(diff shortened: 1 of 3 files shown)


apaskulin (commented):
This looks great! Just a few questions.

- What's the reason behind ignoring headings within a platform context?
- I think we should remove the hover text ("Permanent link"). I haven't seen this on any other sites, and I think the link icon is a well-understood convention.
- I see you've duplicated the footer_js partial from the theme directory. Do we need to keep both files?

rnevius:
@apaskulin Answers to your questions:

1. This seems to be the current pattern in the docs. A good example is the Sensu Go [installation page](https://docs.sensu.io/sensu-go/5.6/installation/install-sensu/). "Waypoints" (the plugin responsible for updating the URL on scroll) are only added if the page doesn't have platform-specific content. The reason is that linking to content that is hidden on page load would mean that the permalink wouldn't work anyway. Please let me know if that doesn't make sense. Here is a similar check [in the `master` branch](https://github.com/sensu/sensu-docs/blob/fa4a54ea24b6e32df55e27bd646499df705c21f2/layouts/_default/single.html#L65-L68), which inspired this decision.
2. As requested, I have removed the `title` attribute from anchor elements.
3. I admit that I'm fairly confused by the organization of this repository when it comes to core theme template modification. Best practices dictate that the base theme's content should never be directly modified (i.e. anything in the `themes/` directory) and Hugo provides [template lookup rules](https://gohugo.io/templates/lookup-order/) to allow overriding templates while still allowing the core theme to be updated. I see a mixture of both in this repo – some core templates have been directly modified, and others have been duplicated into the `layouts/` directory and modified there. 

    Based on Hugo's own recommendations, it seems to make the most sense to move this file into `layouts/` so that it's crystal clear than the template there is not what was included with the theme. Again, if this isn't desired, I can just rewrite the core theme and delete the duplicated file.

Please let me know if you have any additional concerns.

apaskulin:
Thanks for your responses, @rnevius! This is all really helpful to know.

1. That makes total sense, thanks for the background. However, I think the benefit of having anchors on the page outweighs the risk of a broken anchor, especially since the default behavior is to show content for all platforms. If it's alright with you, can you remove the conditional for the platform content? We can see if this causes any issues and re-add it later on if necessary.
2. Thanks! Confirmed on staging.
3. Good to know! Thanks for clarifying. If it's safe to do so, I'm in support of removing the version of footer_js.html from the themes directory to have only one copy of this code in the project.

rnevius:
@apaskulin I have removed the "platform" check from the template. Any heading with an `id` attribute will now include a clickable anchor. 👍 

Regarding removing the `footer_js.html` template from the hugo-material-docs theme, I'm not sure that's necessary or recommended. Currently, any theme layout that has been overridden has been copied from `themes/hugo-material-docs/layouts/` to `layouts/` and has been modified:

<pre><code><b>~/www/sensu-docs $ ls layouts</b>
404.html
_default
index.html
partials
shortcodes

<b>~/www/sensu-docs $ ls themes/hugo-material-docs/layouts/</b>
404.html
_default
index.html
partials
shortcodes</code></pre>

And for "partials" specifically:
<pre><code><b>~/www/sensu-docs $ ls layouts/partials/</b>
alert.html
alertSection.html
drawer.html
footer.html
footer_js.html
head.html
header.html
nav.html
pagination.html
platformDropdown.html
productVersionDrawer.html
productVersionDropdown.html

<b>~/www/sensu-docs $ ls themes/hugo-material-docs/layouts/partials/</b>
drawer.html
footer.html
footer_js.html
head.html
header.html
nav.html
nav_link.html</code></pre>

Obviously, I'm not the repo/product owner; so I can only make a recommendation. If this needs to be removed for some reason, I'll take care of it.

apaskulin (approved):
Deployed to staging and confirmed. Thank you so much, @rnevius!

PR #1425, opened 2019-05-02
Correct production filter examples

### LINKED ISSUE #1408: Filter examples are not possible, should be usable
The examples for certain filters are not possible to use.

## This is an issue with:

## Expected Behavior

Examples in documentation should be relevant and applicable.
## Current Behavior

In reviewing the examples for filters, I found that there are two examples that just wouldn't work, the [Handling production events](https://docs.sensu.io/sensu-go/5.5/reference/filters/#handling-production-events) and [Handling non-production events](https://docs.sensu.io/sensu-go/5.5/reference/filters/#handling-non-production-events).

The reason I say they won't work is because they are filtering on namespaces, while being created in the default namespace.  Since filters and handlers are namespace specific, a filter created in a namespace (such as default in the examples), would never be called upon to filter an event and therefore an entity from a different namespace.
## Possible Solution

Should those examples be using labels, possibly?

1.
2.
3.

## Context

* Browser(s) and version(s): 
* OS:

apaskulin:
## Description

Corrects the production and non-production filter examples to use labels instead of namespaces.

Tested using Sensu Go 5.6 and CentOS7:

- Allow and deny filters work as expected
- Handles case in which there are no entity labels without error
- Adding `event.entity.labels.indexOf('foo') >= 0 ` to the filter expression resulted in ` ..."TypeError: 'indexOf' is not a function"...`

To do:

- [x] Copy to all versions

## Motivation and Context

Closes #1408


content/sensu-go/5.6/reference/filters.md
```diff
@@ -343,54 +343,44 @@ example      | {{< highlight shell >}} "annotations": {
...
 }
 {{< /highlight >}}
 
 ### Handling non-production events
 
-The following example filter definition, entitled `development_filter`, will
-discard event data with a custom entity definition attribute `"namespace":
-"production"`.
-
+The following filter,`development_filter`, discards events with a custom entity label `"environment": "production"`.
```
rgeniesse: Is the intention of this filter to let events be handled when there is no label at all named `environment`? I tested with an entity that had no labels and events were still handled, as the expression is coming back as false.

Also, the naming is a little misleading as `development_filter`. It led me to believe that the label had to be `development`. In it's current state it is just "not production" and it will be handled.

apaskulin: The intention is to allow only events that do not have the `environment: production` label, so events without an `environment` label and events with `environment` set to something other than `production` would be allowed.

+1 to the name change

rgeniesse: Sounds good to me! 
```suggestion
The following filter,`not_production`, discards events with a custom entity label `"environment": "production"`.
```

content/sensu-go/5.6/reference/filters.md
```diff
@@ -343,54 +343,44 @@ example      | {{< highlight shell >}} "annotations": {
...
 Note that `action` is `deny`, making this an exclusive filter; if evaluation
-returns false, the event will be handled.
+returns false, the event is handled.
+
 {{< highlight json >}}
 {
   "type": "EventFilter",
   "api_version": "core/v2",
   "metadata": {
     "name": "development_filter",
```
rgeniesse:
```suggestion
    "name": "not_production",
```

rgeniesse (approved):
💯

PR #1426, opened 2019-05-03
Changing the link id for bonsai

agm650:
## Description

Previously 2 URL had the same id (25): a reference to metadata, and the bonsai one.
With this commit the bonsai URL id is changed to 29 to avoid conflict

## Motivation and Context

On the check reference page, the link to metadata attribute is now correctly pointing to the metadata-attributes section


apaskulin (approved):
Thank you so much, @agm650! Great catch. I added a commit to copy to to our newest pre-release version, 5.7.

agm650:
Just trying to help when I can and those kind of errors are easy to solve :)

PR #1431, opened 2019-05-04
Fix name of continue token header

ccressent:
## Description

Change the name of the continue header from `X-Sensu-Continue` to `Sensu-Continue`. The latter was only used during development and was changed to `Sensu-Continue` right before the pagination feature got merged and released.

## Motivation and Context

There was a late change in the name of the header that holds the continue token for pagination and I most likely failed to communicate that to you @apaskulin. Sorry!

Addresses https://github.com/sensu/sensu-go/issues/2913.


apaskulin (approved):
Thanks, @ccressent!

Issue #395, opened 2018-05-04
Sidebar spacing issues at small screen widths


jspaleta:
## This is an issue with:

- [X] Bug (site functionality or styling)
- [ ] Errata (fix needed for doc content)
- [ ] New content (guide wanted)

## Current Behavior
When using the GNOME Shell feature that allows me to throw a window to the left or right side of a display... for quick split screen window viewing... the navigation bar menu fails to collapse..and spills over into the sidebar nav.
Tested with both firefox and chrome.

But if I manually resize the window to roughly the same size, the navigation layout collapses like I would expect to avoid poor rendering due to text spill over.

I think osX has a similar feature, probably a good idea to try to test there as well.

## Expected Behavior
The nav menu should collapse when I throw the window to left/right split screeen like it does when I manually resize.

## Possible Solution
There's some sort of window width check that isn't firing to collapse the nav menu. I'm not familiar enough with the rendering code...yet... to suggest a fix. It's weird though.. manual resize of the window works.

## Steps to Reproduce (for bugs)

1. Gnome Shell linux desktop
2. grab browser window with sensu-core landing page open and throw to left side of display using display edge split screen feature
3. watch nav menu spill over into sidebar instead of collapsing.

## Screenshots (if appropriate):
Dual Screen setup:
Right screen has firefox and chrome side by side using Gnome Shell split window feature.

![screenshot from 2018-05-04 09-19-19](https://user-images.githubusercontent.com/39937/39641889-4ac36356-4f7c-11e8-8444-d6b80f798ae4.png)

## Context
The rendering spill over makes it impossible to use the sidebar nav.

* Browser(s) and version(s): 
firefox 54.0 Fedora build
chrome Version 64.0.3282.186 (Official Build) (64-bit)
* OS: 
Fedora 24 (I really need to upgrade)

I need to re-test with latest to make sure this isn't a bug in GNOME Shell interaction with browsers.

jspaleta:
I got it to happen on manual resize.. 
its just a couple of pixels width that causing the problem.
Bad Manual Window Width: 960-970  -> nave menu renders with spill over into sidebar
Below 960 -> nav/sidebnar menu collapse
Above 970 -> nav menu renders without spill over

apaskulin:
Also seeing lack of left-side padding at around 1100px width

![image](https://user-images.githubusercontent.com/11339965/46485481-9737fb00-c7b0-11e8-876a-1b989d75f0ce.png)

[...]

PR #1446, opened 2019-05-10
Update search design

### LINKED ISSUE #1413: Create search box
### Context

Currently, the search bar appears after selecting the magnifying glass icon in the top-right corner. Users then need to select the search bar (or use the tab key) to enter search terms. To help users find the information they need, we'd like to change this design to use a more obvious, one-step search box.

### To do

Create a search box that appears by default on every page with the placeholder text "Search the docs". The search bar should be responsive to the size of the page and not overlap with the version and product dropdowns.

### Design reference

https://puppet.com/docs

rnevius:
## Description
Adds an omnipresent search to the homepage and desktop-width devices. Adds a search to the drawer for mobile devices.

Resolves #1413

## Review Instructions
In order to pull this off, the navigation bar, product/version dropdowns, and drawer were majorly refactored. The organization of the styles are improved, and the HTML has been trimmed down where possible. A side effect is that the mobile menu/drawer is less buggy. However, I think a [rewrite of all CSS](https://github.com/sensu/sensu-docs/issues/395#issuecomment-490775801) is still a good idea.


apaskulin (commented):
This looks awesome! Can you increase the width of the search box? For example:

Screen Shot 2019-05-10 at 9 10 30 AM

rnevius:
@apaskulin updated!

apaskulin (approved):
Perfect, thanks! 🎉

PR #1449, opened 2019-05-10
Document usage limit

apaskulin:
## Description

Adds a note about usage limits in Sensu Go.

## Motivation and Context

In Sensu Go 5.7, most API responses include the entity limit as a header, so I think we should include this language in the docs. Note: Licensed instances show the entity limit as 0.


nikkictl (approved):
💯

Issue #1365, opened 2019-04-15
Add YML examples


apaskulin:
For all code samples in the Sensu Go docs that support both JSON and YML, apply the language switcher shortcode and add a YML example.

Depends on #1364

apaskulin:
The language toggler is now live! You can add alternative languages to code samples using the following syntax:

```
{{< language-toggle >}}

{{< highlight yml >}}
type: Asset
api_version: core/v2
metadata:
  name: check_script
  namespace: default
spec:
  filters: null
  sha512: 4f926bf4328fbad2b9cac873d117f771914f4b837c9c85584c38ccf55a3ef3c2e8d154812246e5dda4a87450576b2c58ad9ab40c9e2edc31b288d066b195b21b
  url: http://example.com/asset.tar.gz
{{< /highlight >}}

{{< highlight json >}}
{
  "type": "Asset",
  "api_version": "core/v2",
  "metadata": {
    "name": "check_script",
    "namespace": "default"
  },
  "spec": {
    "url": "http://example.com/asset.tar.gz",
    "sha512": "4f926bf4328fbad2b9cac873d117f771914f4b837c9c85584c38ccf55a3ef3c2e8d154812246e5dda4a87450576b2c58ad9ab40c9e2edc31b288d066b195b21b"
  }
}
{{< /highlight >}}

{{< /language-toggle >}}
```

Questions to resolve:
- Within specification tables, we currently use JSON examples. However, these examples aren't valid JSON, so cannot be labelled with `{{< highlight json >}}`. Possible options include: Using only YML examples in spec tables.

rgeniesse:
I have this working for `*.md` files in any number of directories, with any number of JSON blocks, below the directory holding the Go binary.

Example start and output:

Start with:
```
{{< highlight json >}}
{
  "type": "CheckConfig",
  "api_version": "core/v2",
  "metadata": {
    "name": "interval_check",
    "namespace": "default"
  },
  "spec": {
    "command": "check-cpu.sh -w 75 -c 90",
    "subscriptions": ["system"],
    "handlers": ["slack"],
    "interval": 60,
    "publish": true
  }
}
{{< /highlight >}}
```

Output:

```
{{< language-toggle >}}

{{< highlight yml >}}
type: CheckConfig
api_version: core/v2
metadata:
  name: interval_check
  namespace: default
spec:
  command: check-cpu.sh -w 75 -c 90
  handlers:
  - slack
  interval: 60
  publish: true
[...]

PR #1461, opened 2019-05-15
Have json tags the same everywhere

rgeniesse:
## Description
Noticed during #1365.

## Motivation and Context
To be consistent.

Post change:

```
~/git/sensu-docs/content [master] $ grep -iR '{{< highlight json >}}' . | grep -v '{{< highlight json >}}' | wc -l
       0
```


apaskulin (approved):
⭐️

PR #1462, opened 2019-05-15
Fix Algolia version index

### LINKED ISSUE #1459: Search is not returning any results
Currently, Algolia isn't returning any search results. We're assuming that this regression is a result of #1446. We've confirmed that this isn't an Algolia outage. @rnevius

Algolia DocSearch reference: https://community.algolia.com/docsearch/dropdown.html
Algolia Index: https://github.com/algolia/docsearch-configs/blob/master/configs/sensu.json

rnevius:
Adds the "version" selector to the version dropdown for proper indexing.

Resolves #1459

### DIFF
--- a/layouts/partials/productVersionDropdown.html
+++ b/layouts/partials/productVersionDropdown.html
@@ -22,10 +22,10 @@
   <div class="dropdown dropdown--versions">
     {{ if isset .Params "version" }}
       {{ $display := .Params.version }}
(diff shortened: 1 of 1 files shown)


apaskulin:
Thanks for looking into this, @rnevius! It looks like the Algolia folks have pushed a fix for the current setup here https://github.com/algolia/docsearch-configs/commit/c46053f359ac210572ebae66ea7b06fab1c6ac75 and the search is now working. I'm going to close this PR for now, and we can revisit this later if we run into further issues.

PR #1471, opened 2019-05-16
Document chunk-size flag

### LINKED ISSUE #1467: Add --chunk-size to sensuctl reference
https://github.com/sensu/sensu-go/pull/2966 adds pagination support to `sensuctl` via a `--chunk-size` flag for all the `list` sub-commands.

## This is an issue with:

apaskulin:
## Description

Adds the `chunk-size` flag to the sensuctl reference

## Motivation and Context

Closes #1467

### DIFF
--- a/content/sensu-go/5.8/sensuctl/reference.md
+++ b/content/sensu-go/5.8/sensuctl/reference.md
@@ -16,6 +16,7 @@ menu:
 - [Creating resources](#creating-resources)
 - [Updating resources](#updating-resources)
 - [Managing resources](#managing-resources)
+- [Pagination](#pagination)
 - [Filtering](#filtering) (enterprise-only)
 - [Time formats](#time-formats)
 - [Shell auto-completion](#shell-auto-completion)
@@ -385,6 +386,15 @@ See the [RBAC reference][21] for information about using access control with nam
 #### sensuctl user
 
 See the [RBAC reference][22] for information about local user management with sensuctl.
+
+## Pagination
+
+Sensuctl supports pagination for all `list` commands using the `--chunk-size` flag.
+For example, the following command returns only the first 20 events.
+
+{{< highlight shell >}}
+sensuctl event list --chunk-size 20
+{{< /highlight >}}
 
 ## Filtering
 
--- a/content/sensu-go/5.8/sensuctl/reference.md
+++ b/content/sensu-go/5.8/sensuctl/reference.md
@@ -16,7 +16,6 @@ menu:
 - [Creating resources](#creating-resources)
 - [Updating resources](#updating-resources)
 - [Managing resources](#managing-resources)
-- [Pagination](#pagination)
 - [Filtering](#filtering) (enterprise-only)
 - [Time formats](#time-formats)
 - [Shell auto-completion](#shell-auto-completion)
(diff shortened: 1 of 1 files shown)


content/sensu-go/5.8/sensuctl/reference.md
```diff
@@ -386,6 +387,15 @@ See the [RBAC reference][21] for information about using access control with nam
 
 See the [RBAC reference][22] for information about local user management with sensuctl.
 
+## Pagination
+
+Sensuctl supports pagination for all `list` commands using the `--chunk-size` flag.
+For example, the following command returns only the first 20 events.
```
apaskulin: @ccressent Is there anything else we can say here about the ordering of the objects? Is there a way to see the next X number of objects after the initial command?

ccressent: Sorry Alex, I must have communicated poorly about how this `--chunk-size` flag works.

In your example, `sensuctl event list --chunk-size 20` would retrieve _all_ the events. In fact, purely in term of output, there should never be a difference between `sensuctl x list` and `sensuctl x list --chunk-size y`!

What the flag does is this: instead of making a single query to the API to get all the events in one go, it instead makes a series of smaller queries to retrieve all the events in smaller chunks. For example, if there are 1000 events to be retrieved, and `chunk-size` is 20, `sensuctl` will issue 50 queries to retrieve all the events and then display them.

One might ask: why is that useful? That's because trying to retrieve big sets of objects from an API can lead to issues, such as the query timing out. If instead the client issues many smaller queries and then reassemble the response, most of those problems disappear.

ccressent:
I don't think I would mention "pagination" at all for `sensuctl`. From a `sensuctl` user's point of view,  `chunk-size` is just a way to retrieve objects in smaller chunks rather than all at once. It is mainly there for people trying to retrieve really big list of objects and encounter time outs in the API.

The fact that it uses the API's pagination capabilities internally to achieve that is, in some sense, a detail.

content/sensu-go/5.8/sensuctl/reference.md
```diff
@@ -320,6 +320,15 @@ sensuctl check info check-cpu --format wrapped-json
 In addition to the standard operations, commands may support subcommands or flags that allow you to take special action based on the resource type; the following sections call out those resource-specific operations.
 For a list of subcommands specific to a resource, run `sensuctl TYPE --help`.
 
+### Handling large datasets
+
+When querying sensuctl for large datasets, you use the `--chunk-size` flag with any `list` command to avoid timeouts and improve performance.
```
ccressent: Did you mean "you _can_ use"?

content/sensu-go/5.8/sensuctl/reference.md
```diff
@@ -320,6 +320,15 @@ sensuctl check info check-cpu --format wrapped-json
...
+For example, the following command returns the same output as `sensuctl event list` but makes multiple API queries (each for the number of objects specified by `--chunk-size`) instead of one API query for the complete dataset.
+
+{{< highlight shell >}}
+sensuctl event list --chunk-size 20
```
ccressent: 20 is probably on the small side... I know `kubectl`'s default is 500. What about just going with 500 in the example?

ccressent (approved):

content/sensu-go/5.8/sensuctl/reference.md
apaskulin:
```suggestion
When querying sensuctl for large datasets, you can use the `--chunk-size` flag with any `list` command to avoid timeouts and improve performance.
```

content/sensu-go/5.8/sensuctl/reference.md
apaskulin:
```suggestion
sensuctl event list --chunk-size 500
```

PR #1472, opened 2019-05-16
Document etcd-cipher-suites backend config option

### LINKED ISSUE #1468: Add etcd-cipher-suites backend config option
Page: 5.8 backend reference

To do: Add `etcd-cipher-suites` config flag

Reference: https://github.com/sensu/sensu-go/pull/2956

apaskulin:
## Description

- Updates the backend reference to include the etc-cipher-suites attribute
- Updates the example backend config file
- Updates the securing Sensu guide
- Added defaults based on https://github.com/etcd-io/etcd/blob/master/pkg/tlsutil/cipher_suites.go

To do:
- [x] Add more information in the reference and guide (Reference: https://github.com/sensu/sensu-go/issues/2914)
- [x] Add realistic example
- [x] Open PR for example file in sensu-enterprise-go

## Motivation and Context

Closes #1468

### DIFF
--- a/content/sensu-go/5.8/guides/securing-sensu.md
+++ b/content/sensu-go/5.8/guides/securing-sensu.md
@@ -37,6 +37,8 @@ etcd-peer-cert-file: "/path/to/your/peer/cert"
 etcd-peer-key-file: "/path/to/your/peer/key"
 etcd-peer-client-cert-auth: "true"
 etcd-peer-trusted-ca-file: "/path/to/your/peer/ca/file"
+etcd-cipher-suites
+ - your-cipher
 {{< /highlight >}}
 
 ## Securing the API and the dashboard
--- a/content/sensu-go/5.8/guides/securing-sensu.md
+++ b/content/sensu-go/5.8/guides/securing-sensu.md
@@ -37,8 +37,6 @@ etcd-peer-cert-file: "/path/to/your/peer/cert"
 etcd-peer-key-file: "/path/to/your/peer/key"
 etcd-peer-client-cert-auth: "true"
 etcd-peer-trusted-ca-file: "/path/to/your/peer/ca/file"
-etcd-cipher-suites
- - your-cipher
 {{< /highlight >}}
 
 ## Securing the API and the dashboard
--- a/content/sensu-go/5.8/reference/backend.md
+++ b/content/sensu-go/5.8/reference/backend.md
@@ -657,16 +657,17 @@ no-embed-etcd: true{{< /highlight >}}
 
 | etcd-cipher-suites    |      |
 ------------------------|------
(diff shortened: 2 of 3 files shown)


echlebek (commented):
Should we list the supported ciphers? They can be found here:

https://golang.org/pkg/crypto/tls/#pkg-constants

The ciphers from TLS 1.0 - 1.2 are available, TLS 1.3 ciphers are not available. (But TLS 1.3 can be enabled in another way)

content/sensu-go/5.8/reference/backend.md
```diff
@@ -653,6 +654,22 @@ sensu-backend start --no-embed-etcd
 # /etc/sensu/backend.yml example
 no-embed-etcd: true{{< /highlight >}}
 
+
+| etcd-cipher-suites    |      |
```
apaskulin, line 659: Is there a default value?

content/sensu-go/5.8/guides/securing-sensu.md
```diff
@@ -37,6 +37,8 @@ etcd-peer-cert-file: "/path/to/your/peer/cert"
 etcd-peer-key-file: "/path/to/your/peer/key"
 etcd-peer-client-cert-auth: "true"
 etcd-peer-trusted-ca-file: "/path/to/your/peer/ca/file"
+etcd-cipher-suites
+ - your-cipher
```
apaskulin: What should we include in this guide for this option?

content/sensu-go/5.8/reference/backend.md
echlebek: No - an empty value will lead to the default etcd cipher suites being used. (I'm not sure what they are.)

content/sensu-go/5.8/guides/securing-sensu.md
echlebek: Tough call. This is an advanced option and I wouldn't necessarily want to prescribe something for this.

content/sensu-go/5.8/reference/backend.md
apaskulin: Based on [this doc](https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/security.md): "empty will be auto-populated by Go". So maybe the values listed [here](https://golang.org/pkg/crypto/tls/#pkg-constants) would be the defaults?

cwjohnston (approved):
I think this looks good for the release. It would be ideal if we can identify and document the defaults, but we don't need to be blocked by that.

content/sensu-go/5.8/reference/backend.md
echlebek: That appears to be the case.

https://github.com/etcd-io/etcd/blob/master/embed/config.go#L187

PR #1474, opened 2019-05-17
Correct typo in sensu-go backend example configuration

avanier:
## Description

I found what I think is a typo in the sensu-go example configuration provided in the documentation.

## Motivation and Context

I was about to use this as my configuration file, and it had a typo.

## Review Instructions

- You might not want this backported to older versions of the documentation, your call.
- You might want this to be squashed.


apaskulin (approved):
Great catch, @avanier. Thank you! 🎉

PR #1489, opened 2019-05-27
Remove dashboard filtering document from v5.9

jamesdphillips:
## Description

As of v5.9 filtering has been completely reworked and the old convoluted filter input previously found at the top of list views is no more. As such this wonderful document is no longer relevant. And while I do feel bad about removing this document, we are currently building a new advanced filtering feature that will necessitate the resurrection of this at some point in the future.

For more on the motivation for removing the feature see PR https://github.com/sensu/web/pull/103, an excerpt follows:

> While incredibly flexible the previous Javascript based implementation had a number of irreparable issues.
> * Filtering resources- particularly events and entities- did not scale as amount of agents a cluster monitored increased. This could lead to timeouts, exceptions, and crashes when attemping to filtering hundreds/thousands of resources. The new approach is significantly more efficient and has room for future optimization.
> * Users generally didn't understand how to use the filter box. There were very few affordances to help a user get started, the user needed to have a strong understanding of the shape of Sensu's objects, and required a basic understanding of ECMAScript 5.1. In short- while extremely powerful the user experience was dreadful.
> * Being Javascript based it was very difficult to set the current state of the preset filters. This further made the experience of using lists poor.

## Motivation and Context

No longer relevant.

## Review Instructions

Are there any links / references to this page that I've missed?


apaskulin (approved):
:tada: Thanks, @jamesdphillips! I added a redirect for this page to the dashboard overview and double-checked for any additional links we would need clean up. :shipit:

PR #1495, opened 2019-05-28
Conform the type attribute to not be uppercase

rgeniesse:
## Description
Makes any JSON example that uses the `type` attribute to not be upper case.

## Motivation and Context
While it might be valid to use `Type`, the documentation site uses `type` everywhere else.


apaskulin (approved):
🙌

PR #1502, opened 2019-05-30
Fix version ordering in dropdowns

### LINKED ISSUE #170: Sorting on version pulldown is not quite right
2018-02-05 at 10 12 am

rnevius:
## Description
Config now includes a hash for each version of each product. These hashes are more flexible/powerful and contain `platform` and `version` keys. This is similar to this [idea proposed in #170](https://github.com/sensu/sensu-docs/issues/170#issuecomment-369358882), but addresses a problem with that approach. 

Closes #170 

## Motivation and Context

Versions should be ordered properly in all dropdowns.

## Review Instructions

As a result of the new config, platform content was also affected by these changes. This means that platform dropdowns/shortcodes and version dropdowns should be checked for consistency.


apaskulin (commented):
I’ve deployed this to staging and spot-checked the versions and platform content, and it looks correct! (Thanks for keeping in sync with our two latest releases 💚)

However I’m a bit apprehensive about switching the site config from TOML to JSON. It seems like we’re losing a lot in terms of readability and compactness. Can you elaborate on why this approach is necessary vs. a solution that uses the TOML config?

rnevius:
@apaskulin I have translated the JSON updates back into TOML.

apaskulin (approved):
Awesome work! Thanks, @rnevius! I've deployed this to staging and spot-checked for correctness.

PR #1507, opened 2019-05-31
correct namespace

rgeniesse:
## Description
Corrects a check example that should be using the `production` namespace as described in the text above it.

## Motivation and Context
To be correct!


apaskulin (approved):
🙌

PR #1529, opened 2019-06-10
Fixed a typo

bananabele:
Changed complimenting -> complementing

## Description

Fixed a little typo on this page

## Motivation and Context

See above


apaskulin (approved):
💚 Thanks, @bananabele! I've added a commit to copy the fix to all versions.

PR #1536, opened 2019-06-12
Add troubleshooting section to plugin install docs

### LINKED ISSUE #1521: Need a note about  needing development tools before using sensu-install
sensu-install instractions for disk-checks will fail on freshly provisioned  linux systems.

## This is an issue with:

## Context
On a freshly provisioned Centos VM the instructions found in
https://docs.sensu.io/sensu-go/5.9/installation/plugins/#installing-plugins-using-the-sensu-install-tool

The sensu-install will fail to install the disk-checks plugin because the necessary Developer Tools, such as the C compiler and related executables are not on the system yet.
We should probably add a note that common development tools for the target host will need to be installed, because some of the ruby gem dependencies have native libraries that need to be compiled as part of the install process.

Unfortunately the fix is different for different linux environments.
On Centos and other RPM based linux distros I have resolved this by doing: 
`yum groupinstall "Development Tools"`

On Debian and dpkg based linu distros this would be resolved with a different command. My best guess for the command is this:
` apt-get install build-essential`

This is just one of those things people who work with Ruby gems find out about as they go. And its difficult to know which Ruby gem will end up needed to compile a C library, because of the way depedencies work with Ruby, you sort of don't know until you try to install something and one of its deps won't install.

apaskulin:
## Description

- Replaces the disk-checks plugin in the main instructions with influxdb, which doesn't require dev tools
- Adds a troubleshooting section with commands to install dev tools on supported Linux platforms
- Removes note about handler plugins not being compatible with Sensu Go in favor of guidance in individual plugin READMEs
- Removes typo

Tested using CentOS 7 and Ubuntu 16

To do:

- [ ] Copy to all versions

## Motivation and Context

Closes #1521

### DIFF
--- a/content/sensu-go/5.0/installation/plugins.md
+++ b/content/sensu-go/5.0/installation/plugins.md
@@ -13,7 +13,7 @@ menu:
 Sensu's functionality can be extended through the use of plugins.
 Plugins can provide executables for performing status or metric checks, mutators for changing data to a desired format, or handlers for performing an action on a Sensu event.
 
-### Installing plugins using Bonsai, the Sensu asset index
+## Installing plugins using Bonsai, the Sensu asset index
 
 Assets are shareable, reusable packages that make it easy to deploy Sensu plugins.
 You can use assets to provide the plugins, libraries, and runtimes you need to automate your monitoring workflows.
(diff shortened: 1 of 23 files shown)


rgeniesse (approved):
🎉

PR #1537, opened 2019-06-12
Update change user password command to include required flags.

rgeniesse:
## Description
The current change user password command needs updated to include required flags.

## Motivation and Context
Makes it easier for users to be given the exact command they need to change a users password, including required flags.

Without them, `sensuctl` returns an error.


apaskulin (approved):
Nice catch! Thanks, @rgeniesse!

PR #1552, opened 2019-06-19
Add release notes for Sensu Go 5.10

apaskulin:
## Description

- Adds release notes for Sensu Go 5.10
- Updates the dashboard overview to list entity deletion as free-tier
- Adds license as an RBAC resource type
- Updates the build ID in Windows package names

The following changelog items are omitted from the release notes:

- The resources metadata are now validated with the request URI.
- [Web] Use new filter syntax on homepage; avoids having to filter entire set of events.
- [Web] Updating the embedded assets no longer requires node & yarn.
- [Web] Publish tessen metrics for homepage tile click events.
- Added a SetNamespace(string) function to enterprise resources to satisfy the Resource interface.
- Added POST /api/core/v2/tessen/metrics.
- Added the ability in TessenD to listen for metric points on the message bus, populate, and send them to the Tessen service.
- Added a tag to all Tessen metrics to differentiate internal builds.
- [Web] Updated embedded web assets from 275386a ... 46cd0ee
- Refactoring of the REST API.
- Changed the identifying cluster id in TessenD from the etcd cluster id to the sensu cluster id.
- [GraphQL] Updates PutResource mutation to accept an upsert boolean flag parameter. The upsert param defaults to true, but if set to false the mutation will return an error when attempting to create a duplicate resource.
- Eventd has been refactored. Users should not perceive any changes, but a substantial amount of business logic has been moved into other packages.
- Eventd now uses a constant number of requests to etcd when working with silenced entries, instead of a number that is proportional to the number of subscriptions in a check.

### DIFF
--- a/content/sensu-go/5.10/dashboard/overview.md
+++ b/content/sensu-go/5.10/dashboard/overview.md
@@ -54,8 +54,6 @@ You can create, edit, and delete Sensu checks using the dashboard checks page.
 
 ### Managing entities
 
-**LICENSED TIER**: Unlock entity management in the Sensu Go dashboard with a Sensu license. To activate your license, see the [getting started guide][6].
-
 You can delete Sensu entities using the dashboard entities page.
 
 ### Managing handlers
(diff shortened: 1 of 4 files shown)


jamesdphillips (approved):
Change to RBAC docs looks good to me :D

wesleycyu (approved):

PR #1561, opened 2019-06-24
Update etcd cipher suites

palourde:
Signed-off-by: Simon Plourde <simon@sensu.io>

Background: https://github.com/sensu/sensu-go/issues/3019

The list of cipher suites used by etcd appear to be sensitive to the declaration order and the process will mysteriously fail with the default list of ciphers provided in our documentation.

Considering the default cipher list and ordering depends on the hardware used, I'd suggest we provide a recommended list instead, which has been successfully tested.


apaskulin (approved):
⭐️ Thanks, @palourde!

PR #1576, opened 2019-06-28
Refactor site styles

### LINKED ISSUE #395: Sidebar spacing issues at small screen widths

### LINKED ISSUE #1528: Refactor site styles
- Add CSS pre-processing to the build process – We currently have to add vendor prefixes to some style definitions manually, which is unnecessary and vulnerable to human error.
- Split stylesheets into components to keep things better organized – this will make updating/adding new styles easier in the future
- Standardize media queries to be mobile-first and use sensible breakpoints – some styles are currently desktop-first and others are mobile-first which has led to a lot of confusing behavior when components with different layout strategies are nested
- Refactor the HTML layouts (and corresponding CSS definitions) so that we're using fewer one-off ID attributes in favor of reusable class attributes
- Merge these changes and verify that the site functionality/appearance hasn't changed (aside from bugs being fixed)

rnevius:
## Description

- Adds SCSS/CSS pre-processing to the Grunt build process. SCSS is compiled to browser-prefixed and minified CSS. In development, all changes are automatically recompiled and reloaded. As a result, built CSS is no longer tracked by this repository.
- Browser breakpoints are standardized and styles are mobile-first
- Random bug fixes have been applied (sidebar issues, `<table>` rendering issues, etc.)

The new styles reduce the CSS payload in production by 80% and are contained in a single file. Mostly, a lot of unnecessary HTML markup and attributes have been removed and simplified.

## Motivation and Context

Styles should be properly defined before adding new styled components.

Closes #1528 
Closes #395

## Review Instructions

The Docs site should be built in staging and tested on mobile/desktop environments. If a bug is discovered, please cross-reference with the current production build. If a bug exists in both production and this PR, it should probably be opened as a new issue (separate from this PR). If the bug only exists in this PR, it should be fixed before merging into `master`.


apaskulin (commented):
Thanks, @rnevius! I've deployed this to staging. The only thing I've noticed so far is that, in the sidebar, the `v` icons briefly appear as squares during page load.

rnevius:
@apaskulin FontAwesome icon flashing is now fixed.

apaskulin (approved):
Thanks, @rnevius! I've confirmed the fix for the flashing icons. Everything on staging looks good! Barring any additional issues, I'll plan to merge this on Monday.

PR #1627, opened 2019-07-23
Remove versions page

apaskulin:
## Description

The versions page adds significant complexity to the versioning process and is largely redundant with the release notes and Sensu terms and conditions. This PR removes it from the site.

## Motivation and Context

From May 1 - July 23, there have been 219 unique pageviews to the versions page, with 30% of visitors accessing the 5.7 version. We can compare this with 1,582 unique pageviews for the release notes over the same period.


rgeniesse (approved):
Agreed! :shipit:

Issue #1586, opened 2019-07-08
Resolve security alerts


apaskulin:
Context: https://github.com/sensu/sensu-docs/network/alerts

### To do

Investigate and resolve security alerts for:

- [underscore.string](https://github.com/sensu/sensu-docs/network/alert/yarn.lock/underscore.string/open)
- [js-yaml](https://github.com/sensu/sensu-docs/network/alert/yarn.lock/js-yaml/open)
- [lodash](https://github.com/sensu/sensu-docs/network/alert/yarn.lock/lodash/open)
- [string](https://github.com/sensu/sensu-docs/network/alert/yarn.lock/string/open)

cwjohnston:
All alerts are now resolved ✅

PR #1632, opened 2019-07-24
update js-yaml dependency to address security alert

cwjohnston:
## Description

Updates js-yaml module to ^3.13.1

## Motivation and Context

Addresses one alert described in #1586


cwjohnston:
Closing in favor of #1633 which removes this dependency.

PR #1633, opened 2019-07-24
Remove lunr search indexing

cwjohnston:
### Description

Removes grunt task and related dependencies for generating search index via Lunr

## Motivation and Context

We now use Algolia for search, so Lunr is no longer needed.

Also removes `string` and `js-yaml` javascript modules for which we have some open security alerts in #1586 .

## Review Instructions

I'll deploy to stage and we can confirm if search still works.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,6 +1,4 @@
 const sass = require('node-sass');
-var yaml = require("js-yaml");
-var S = require("string");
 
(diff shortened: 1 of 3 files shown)


cwjohnston:
Search still works in staging, but of course the results link to production 🙃 

I think this is good to merge.

apaskulin (approved):
Reviewed on staging and everything looks good!

PR #1652, opened 2019-08-02
add index function syntax examples for labels and annotations

### LINKED ISSUE #1651: Update Sensu Go Token Substitution with examples using index function to access labels and annotations with "dots" in the name.
Provide note in token substitution reference on how to access labels and annotations using the index function fro go template.

corner case in golang template operation. "dots"  get intepreted as object seperators.
an annotation key like "this.example" gets intepretted incorrectly using standard syntax

{{annotations.this.example}}   fails because "this" is interpreted as an object 
What we need instead is to use construct like

{{index annotations "this.example"}}  would succeed because we've explicit asked the template engine to treat annotations as a key/value mapping and we are asking for the value for the "this.example" key and explicitly indicating the key is a string.

annotations and labels with dots in the name break your brain. Need to breadcrumb correct syntax when we introduce token subs to inception the index function for people to vaguely remember so they can go try to look up how to use it when they run into trouble.

## This is an issue with:

jspaleta:
## Description
Add token sub index function example to handle labels/annotations with dots in the name

## Motivation and Context
dots are allow in key names.. but standard token sub template syntax is confused by them. 
Need to add example of using the index function as an aid for people to find.

this Fixes #1651


content/sensu-go/5.11/reference/tokens.md
```diff
@@ -36,7 +36,11 @@ Tokens are invoked by wrapping references to entity attributes and labels with d
 - `{{ .name }}` would be replaced with the [entity `name` attribute][3]
 - `{{ .labels.url }}` would be replaced with a custom label called `url`
 - `{{ .labels.disk_warning }}` would be replaced with a custom label called
+- `{{ index .labels "disk_warning" }}` would be replaced with a custom label called
   `disk_warning`
+- `{{ index .labels "cpu.threshold" }}` would be replaced with a custom label called `cpu.threshold`
+
+Note: when an annotation or label name has a dot (ex: `cpu.threshold`), the template index function syntax must be used to ensure correct processing, as the dot notation is also used for object nesting.
```
apaskulin:
```suggestion
_NOTE: when an annotation or label name has a dot (ex: `cpu.threshold`), the template index function syntax must be used to ensure correct processing, as the dot notation is also used for object nesting._
```

content/sensu-go/5.12/reference/tokens.md
```diff
@@ -36,7 +36,11 @@ Tokens are invoked by wrapping references to entity attributes and labels with d
 - `{{ .name }}` would be replaced with the [entity `name` attribute][3]
 - `{{ .labels.url }}` would be replaced with a custom label called `url`
 - `{{ .labels.disk_warning }}` would be replaced with a custom label called
+- `{{ index .labels "disk_warning" }}` would be replaced with a custom label called
   `disk_warning`
+- `{{ index .labels "cpu.threshold" }}` would be replaced with a custom label called `cpu.threshold`
+
+Note: when an annotation or label name has a dot (ex: `cpu.threshold`), the template index function syntax must be used to ensure correct processing, as the dot notation is also used for object nesting.
```
apaskulin:
```suggestion
_NOTE: when an annotation or label name has a dot (ex: `cpu.threshold`), the template index function syntax must be used to ensure correct processing, as the dot notation is also used for object nesting._
```

apaskulin (approved):
Looks good! I Added suggestions to apply the NOTE callout style.

content/sensu-go/5.11/reference/tokens.md
```diff
@@ -36,7 +36,11 @@ Tokens are invoked by wrapping references to entity attributes and labels with d
 - `{{ .name }}` would be replaced with the [entity `name` attribute][3]
 - `{{ .labels.url }}` would be replaced with a custom label called `url`
```
cwjohnston, line 37: @jspaleta seems like lines 37 and 38 should be removed as they are examples of the usage this PR seeks to update. WDYT?

apaskulin: From jef: "I want to point out both syntaxes as they will see both in practice when looking at configs in the wild. Index notation is only needed if the key string includes a dot."

PR #1654, opened 2019-08-02
use explicit javascript dictionary syntax for label indexing in filter examples

jspaleta:
## Description
change label filter examples with dots to use javascript dictionary index syntax.

## Motivation and Context
labels/annotations with "dots" in the name need to use explicit index syntax

this works:
```
event.check.labels["this.example.msg"] == "hey"
```

this fails
```
event.check.labels.this.example.msg == "hey"
```


rgeniesse:
@jspaleta can you fix the JSON issues called out in https://travis-ci.com/sensu/sensu-docs/jobs/222153325? Thanks!

jspaleta:
ah I fell into the json trap!.. I tested the updated filter syntax using a live javascript editor, but the configuration its embedded in not valid json.....

rgeniesse (approved):
LGTM :shipit:

PR #1659, opened 2019-08-05
Move sensu-backend installation to before sensuctl

rgeniesse:
## Description
The sensu-backend needs to be installed before sensuctl, as sensuctl requires the sensu-backend to be available to properly configure it.

You could install sensuctl, do sensu-backend, then configure sensuctl, but this seems less jumbled.

## Motivation and Context
If used in the current order when configuring sensuctl, it will fail to find the backend running as the backend has not been installed or configured yet

## Review Instructions
- [ ] Copy to version 5.12 (5.10 and lower have the older install style page)


apaskulin (commented):
+1 to copy to 5.12! Thanks!

PR #1669, opened 2019-08-13
Update Enterprise migration guide integration table

cwjohnston:
## Description

* Remove SNMP from integrations table
* Remove Ruby-based asset from graphite entry in table

## Motivation and Context

The linked Sensu SNMP plugins do not include a handler for sending SNMP traps, therefore isn't an analog for the Enterprise SNMP integration.

The asachs01/sensu-plugins-graphite asset does include some ruby-based handlers but I think we should direct folks to the nixwiz/sensu-go-graphite-handler more prescriptively.


asachs01 (approved):
👍 LGTM

PR #1747, opened 2019-09-10
Fix token substitution examples in 1.x migration guide

cwjohnston:
## Description

Fixes examples in the 1.x migration guide to show the correct double curly-brace syntax for token substitution.

## Motivation and Context

Accuracy :)


Blue0ctober (approved):
LGTM

PR #1822, opened 2019-10-08
Add group prefix example and table entry

### LINKED ISSUE #1321: RBAC docs need to detail use of auth (AD/LDAP) prefixes and provide example usage
## This is an issue with:

## Expected Behavior

Documentation for Role Binding and Cluster Role Binding should include the configurations necessary for working with auth providers such as LDAP/AD.  The specification for the name attribute in the [subjects specification](https://docs.sensu.io/sensu-go/5.3/reference/rbac/#subjects-specification) does not mention user or group prefixes.  Also, there should be examples that match these possible configurations.  These are alluded to in the quote below, but do not exist.

> Make sure to include the group prefix and username prefix when creating Sensu role bindings and cluster role bindings. See the RBAC reference for more information about configuring permissions in Sensu and implementation examples.

In fact, prefixes are not mentioned at all in the [RBAC reference](https://docs.sensu.io/sensu-go/5.3/reference/rbac/).

1.
2.
3.

## Context

I am currently doing a PoC for Sensu Enterprise (Go) and have AD auth working, but struggling to proper authorization via a cluster-role-binding for group members.

* Browser(s) and version(s): 
* OS:

rgeniesse:
## Description
Adds a table entry and example for using group prefixes.

## Motivation and Context
Closes #1321 

## Review Instructions
- [x] Not likely the 2nd example table entry, but not sure what else to do here to call it on in reference.
- [x] Just a role binding example?
- [x] Just a group_prefix example?


content/sensu-go/5.13/reference/rbac.md
```diff
@@ -808,6 +808,7 @@ description  | Username or group name.
 required     | true 
 type         | String
 example      | {{< highlight shell >}}"name": "alice"{{< /highlight >}}
+example with prefix | {{< highlight shell >}}"name": "ladp:alice"{{< /highlight >}}
```
hillaryfraley:
```suggestion
example with prefix | {{< highlight shell >}}"name": "ad:alice"{{< /highlight >}}
```

hillaryfraley: Just suggest changing to `ad` to be consistent with the group prefix examples you give later.

content/sensu-go/5.13/reference/rbac.md
```diff
@@ -1179,3 +1224,5 @@ You can add these resources to Sensu using [`sensuctl create`][31].
 [35]: #cluster-role-bindings
 [36]: ../../sensuctl/reference#creating-resources-across-namespaces
 [37]: ../license
+[38]: ../../installation/auth/#groups-prefix
```
hillaryfraley, line 1227: This links to the `groups_prefix` in the LDAP spec (https://docs.sensu.io/sensu-go/5.14/installation/auth/#spec-attributes)

Since this example uses AD, I think we would need to add something like `<a name="ad-groups-prefix"></a>` just before the AD `groups-prefix` description in https://docs.sensu.io/sensu-go/5.14/installation/auth/#active-directory-spec-attributes, right?

I guess we could also revise the role/role binding with groups prefix examples section to use LDAP rather than AD, but I assume there is a good reason you used AD there. What do you think?

content/sensu-go/5.13/reference/rbac.md
```diff
@@ -943,6 +944,50 @@ The following role and role binding give a `dev` group access to create and mana
 }
 {{< /highlight >}}
 
+### Role and role binding examples with a group prefix
+
+If a [groups prefix][38] of `ad` has been configured for [Active Directory authentication][39], the following role and role binding give a `dev` group access to create and manage Sensu workflows within the `default` namespace.
```
hillaryfraley:
```suggestion
In the following code example, if a [groups prefix][38] of `ad` is configured for [Active Directory authentication][39], this role and role binding will give a `dev` group access to create and manage Sensu workflows within the `default` namespace.
```

hillaryfraley (changes requested):
@rgeniesse I added a few comments -- please let me know if anything doesn't make sense in them.

content/sensu-go/5.13/reference/rbac.md
rgeniesse: There seems to already be a `<a name="ad-groups-prefix"></a>` before the groups prefix. So https://docs.sensu.io/sensu-go/5.14/installation/auth/#ad-groups-prefix works already!

I'll make the change so we use AD everywhere.

hillaryfraley (approved):
Now it LGTM!

PR #1825, opened 2019-10-08
Update build numbers for 5.14

hillaryfraley:
## Description
Updates `install-sensu` page for release 5.14

### DIFF
--- a/content/sensu-go/5.14/installation/install-sensu.md
+++ b/content/sensu-go/5.14/installation/install-sensu.md
@@ -212,15 +212,15 @@ sensuctl configure -n \
 --url 'http://127.0.0.1:8080'
 {{< /highlight >}}
 
-Here the `-n` flag triggers non-interactive mode.
-Run `sensuctl config view` to see your user profile.
+Here the `-n` flag triggers non-interactive mode. Run `sensuctl config view` to see your user profile.
+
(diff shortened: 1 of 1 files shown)


content/sensu-go/5.14/installation/install-sensu.md
```diff
@@ -212,15 +212,15 @@ sensuctl configure -n \
 --url 'http://127.0.0.1:8080'
 {{< /highlight >}}
 
-Here the `-n` flag triggers non-interactive mode.
-Run `sensuctl config view` to see your user profile.
+Here the `-n` flag triggers non-interactive mode. Run `sensuctl config view` to see your user profile.
```
rgeniesse, line 215: Were these new line changes intention? If so, is this something we want to put to the rest of the install sensu pages later?

hillaryfraley: Oh -- the single hard return causes words/sentences to run together in the PDF conversion -- the return is removed and it would look like "...mode.Run..." in the PDF. I noticed it here so I removed it (but I can put it back if it's a problem).

I'm not sure if it will end up being a doc-wide backward-looking chore or just something we fix going forward.

rgeniesse: Ah, the joys. Yeah, we can decide that later. Thanks!

rgeniesse (approved):
💯

PR #1850, opened 2019-10-16
Add `list` to resource_names description

### LINKED ISSUE #1847: RBAC reference errata: resource_names
Per a customer ticket, it was noted that there is an inaccuracy in the RBAC reference for `resource_names` and the verbs that it can be used with. `list` needs to be added to this (L#553)

hillaryfraley:
## Description
Add `list` to `resource_names` description

## Motivation and Context
Closes #1847


hillaryfraley:
@asachs01 , this was the only place I saw `list` missing as a verb -- is this correct?

If it is, should this be propagated in the docs for all Sensu Go versions?

asachs01 (approved):

PR #1861, opened 2019-10-22
Replace downloads page links with links for individual platform packages

hillaryfraley:
Jef noted on Slack:
"we seems to have dropped most of our download links for Sensu Core when transitioned to the new website.  The Sensu Core docs point to a downloads page on our main site that no longer exists. The linux instructions in the Sensu Core docs still work, but the exotic platforms like AIX, Win, OSX, etc. no longer have discoverable download links.
I'm assuming this was just an oops on our part, effectively EOL Core a couple of months early.
I've put the links in discourse in the context of the discussion,  but should we drive a change into the Sensu Core docs as well and point people to the correct repository.sensuapp.org directories for each platform?"

## Description
Replaced https://sensuapp.org/download and https://sensuapp.org/downloads links with individual platform links listed at https://discourse.sensu.io/t/sensu-core-download-all-the-versions/1406/2


asachs01 (approved):
Testing links looks good to me. 👍

PR #1866, opened 2019-10-23
Add documentation for agent-auth & asymmetrical jwts

### LINKED ISSUE #1860: Document the new config attributes for asymmetric JWTs
JSON Web Tokens (JWTs) will soon be able to be signed using asymmetric cryptography; see https://github.com/sensu/sensu-enterprise-go/issues/689.

We need to document the new config attributes in sensu-backend and the expected behaviors around existing JWTs and using only the public key.

palourde:
Signed-off-by: Simon Plourde <simon@sensu.io>

## Description
I added documentation around two new configuration attributes in sensu-backend that will be part of the upcoming release.

Long story short, these two new attributes allow users to specify a private and public keys to use to sign the JWTs, which are essentially the tokens provided to users when authenticating. This will be required for users of federation, so the same client can talk to different clusters with the same access token.

Additionally, I updated the `sensu-backend start --help` usage with actual output and realized we missed a couple of `agent-auth-*` attributes, so I also documented them.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/1860


hillaryfraley (approved):

hillaryfraley:
@palourde Thank you! I made some minor proofreading changes and otherwise just reorganized the flag descriptions so they were listed in the same order as in the configuration summary a user would see with `sensu-backend start --help`

Based on your note with this PR, it sounds like I need to add these four flags and descriptions in previous docs versions:
- agent-auth-cert-file string
- agent-auth-crl-urls strings
- agent-auth-key-file string
- agent-auth-trusted-ca-file string

I looked back through the release notes, but I don't see these mentioned as something new. Does that mean they belong in /reference/backend/ all the way back to Sensu Go 5.0?

palourde:
@hillaryfraley Thanks! 

Good catch on the previous docs versions; these flags were actually introduced in the [5.14.0](https://github.com/sensu/sensu-enterprise-go/blob/master/CHANGELOG.md#5140---2019-10-08) release.

hillaryfraley:
@palourde Oh good, then only one version to update! I will create a separate PR to handle it. Thank you!

PR #1867, opened 2019-10-23
Propagate sandbox changes to all versions

hillaryfraley:
Propagate changes from https://github.com/sensu/sensu-docs/pull/1833 through all docs versions


cwjohnston (approved):
I've checked for the presence of front matter changes and found none. Let's do it 👍

PR #1871, opened 2019-10-24
Fix RBAC reference docs around resource_names

palourde:
This PR fixes the reference documentation for the RBAC `resource_names` attribute; it falsely stated that this attribute could be used with requests using the `list` verb.

I also tried to change the language around it to make it more clear that you can still define rules that use other verbs but it will simply not take into consideration the `resource_names` list for these requests.

### DIFF
--- a/content/sensu-go/5.0/reference/rbac.md
+++ b/content/sensu-go/5.0/reference/rbac.md
@@ -535,7 +535,7 @@ example      | {{< highlight shell >}}"resources": ["checks"]{{< /highlight >}}
 
 resource_names    | 
 -------------|------ 
-description  | Specific resource names that the rule has permission to access. Resource name permissions are only available for `get`, `list`, `update`, and `delete` verbs.
+description  | Specific resource names that the rule has permission to access. Resource name permissions are only taken into account for requests using `get`, `update`, and `delete` verbs.
 required     | false
 type         | Array
 example      | {{< highlight shell >}}"resource_names": ["check-cpu"]{{< /highlight >}}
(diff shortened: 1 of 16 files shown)


hillaryfraley:
Disregard -- Cameron explained it to me

@palourde should we also remove it in the Role and Cluster Role examples? e.g.:

```
type: Role
api_version: core/v2
metadata:
  name: namespaced-resources-all-verbs
  namespace: default
spec:
  rules:
  - resource_names: []
    resources:
    - assets
    - checks
    - entities
    - events
    - filters
    - handlers
    - hooks
    - mutators
    - rolebindings
    - roles
    - silenced
    verbs:
    - get
    - list
    - create
    - update
    - delete
```

hillaryfraley:
From @palourde 's explanation in Slack:

Using a RBAC role, I could only grant a user the ability to view a given check (here check_cpu) using the following rule:
```
rules:
  - resource_names:
    - check_cpu
    resources:
    - checks
    verbs:
    - get
```
Then, the user could do `sensuctl check info check_cpu`.
Now, if a user wants to list all checks, he will need more permissions than that right? So I would have to add the following rule:
```
rules:
  - resources:
    - checks
    verbs:
    - list
```
The important thing to note here is that the verb `list`, just like `create`, does not support the `resource_names` attribute, meaning it will ignore whatever is configured within this array. This is what I’m trying to signify in the documentation PR. There’s a couple of technical reasons why `list` and `create` does not support `resource_names`, we can talk about it and possible solutions/workaround but it doesn’t change the current situation.

Regarding the dashboard, the main thing it does when you access a view is listing a given resource type. That’s why if you want to use the dashboard, being able to list certain resources is basically a requirement. (edited) 

In summary, if you want to use all functionalities of the dashboard, you need list permissions on all resources within that namespace.

And no, you can’t filter resources when you list them by specifying a resource_name array in a rule (edited)

hillaryfraley:
After discussing w/ @cwjohnston , this PR is 👍 and I'll approve/merge it.

hillaryfraley (approved):

PR #1874, opened 2019-10-25
Archive docs versions 5.0 through 5.11

hillaryfraley:
Testing out instructions from @cwjohnston 

## Description
I drafted instructions based on Cameron's notes from yesterday: https://docs.google.com/document/d/1WLAEITdQj8wagXH9rrvJ3DhEfRtRLzaOnHa1FZeJHR8/edit#

If this is all correct, I can add the instructions to the docs wiki.

## Motivation and Context
We need to establish the process for archiving docs versions while keeping the archived versions available so we can create new offline PDF copies if needed.


hillaryfraley:
I think the offline docs `yarn run server --layoutDir=offline` is specific to /content/sensu-go? I can't get the archived docs to load in my browser from http://localhost:1313/archived/sensu-go/5.0/ (so I can test creating a PDF).

Edit -- Ok jeez, never mind! I figured it out -- still there under http://localhost:1313/sensu-go/5.0/

content/sensu-go/5.0
```diff
@@ -0,0 +1 @@
+../../archived/sensu-go/5.0
```
cwjohnston, line 1: @hillaryfraley we don't want to commit these symlinks, just create them locally on an as-needed basis (e.g. create symlink so we can generate PDF of archived docs version from server running on local laptop)

hillaryfraley: Ahhhh ok so we stop and commit after `Edit config.toml to remove the version being archived`?

hillaryfraley:
Superseded by https://github.com/sensu/sensu-docs/pull/1875

PR #1885, opened 2019-10-29
Add token substitution example in hooks reference

### LINKED ISSUE #1881: add token substitution example in hooks reference.
We should have an explicit example of a check hook making use of a token substitution and link back to the token substitution reference.

In it we should remind users that token substitution uses entity scoped metadata, so labels should be set at the entity level.

Reference:
https://discourse.sensu.io/t/sensu-hooks-passing-parameters/1419

## This is an issue with:

1.
2.
3.

## Context

* Browser(s) and version(s): 
* OS:

hillaryfraley:
## Description
Adds example of a check hook using token substitution

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/1881

## Review Instructions
I made my best guess at the yml/json examples, but these definitely need a hand.


cwjohnston (approved):
It would be good if we can think of a more concrete example, but for lack of one coming to mind, let's go ahead with this needed addition.

PR #1964, opened 2019-11-27
Replace check example w/example that includes hook in token reference

### LINKED ISSUE #1884: Update token substitution reference to include mention of hooks
Update text of token substitution reference to include discussion about tokens in check hooks.
Right now the section only mentions checks but we can use token sub in hooks as well.

We should be able to extend the existing check example to include a hook, both making use of entity scoped labels for subsitution

## This is an issue with:

1.
2.
3.

## Context

* Browser(s) and version(s): 
* OS:

hillaryfraley:
## Description
Replaces check example with an example that includes a hook

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/1884

### DIFF
--- a/content/sensu-go/5.0/reference/tokens.md
+++ b/content/sensu-go/5.0/reference/tokens.md
@@ -71,8 +71,45 @@ For example, token substitution **cannot** be used for specifying a check interv
 
 ### Token substitution for check thresholds 
 
-In this example [check configuration][5], the `check-disk-usage.go` command accepts `-w` (warning) and `-c` (critical)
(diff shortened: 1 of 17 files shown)


hillaryfraley:
This doesn't like the JSON -- I'll take another look at it tomorrow.

hillaryfraley:
@jspaleta I think the JSON example is causing the doc site build to fail. I've tried to fix it -- I narrowed it down to the close-brace/open-brace between the hook object and the check object, but I cannot figure out how to fix it.

Would you actually create the objects separately? If so, I could put them in separate code formatting blocks. But I don't want to make this change if you actually create them all at once.

hillaryfraley:
I separated the hook and check code so that the JSON would pass the sensu-docs build checks.

Issue #1621, opened 2019-07-22
Update events API docs


apaskulin:
Reference: https://github.com/sensu/sensu-go/pull/3155

To do:

- [ ] Update events API docs
- [ ] Update all sandbox lessons and Katacoda scenarios

hillaryfraley:
I think this is done via https://github.com/sensu/sensu-docs/issues/1680?

hillaryfraley:
- [x] Test removal of namespace in required example under https://docs.sensu.io/sensu-go/5.15/api/events/#eventsentitycheck-put-parameters
- [x] If test returns as expected (example returns response that includes namespace 'default'), remove namespace in required example under https://docs.sensu.io/sensu-go/5.15/api/events/#eventsentitycheck-put-parameters
- [ ] If test returns as expected, remove namespace in step 2, 3, and 4 of sandbox Katacoda scenario and re-test scenario

hillaryfraley:
See https://github.com/sensu/katacoda-scenarios/pull/56

hillaryfraley:
PR in progress: https://github.com/sensu/sensu-docs/pull/1995

PR #1995, opened 2019-12-06
Remove namespace from required minimum example

hillaryfraley:
## Description
Remove namespace from required minimum example under https://docs.sensu.io/sensu-go/5.15/api/events/#eventsentitycheck-put-parameters.

I tested this call by sending the required minimum example without a namespace included in the body of the request. The entity was created successfully without the namespace, so the namespace is not actually required.

See comment below -- I'm not sure it's correct that you can exclude the namespace in all instances. It didn't work in katacoda (where we configure sensuctl with a namespace).

## Motivation and Context
https://github.com/sensu/sensu-docs/issues/1621

### DIFF
--- a/content/sensu-go/5.0/api/events.md
+++ b/content/sensu-go/5.0/api/events.md
@@ -570,8 +570,7 @@ curl -X PUT \
   "entity": {
     "entity_class": "proxy",
     "metadata": {
-      "name": "server1",
-      "namespace": "default"
+      "name": "server1"
     }
   },
   "check": {
--- a/content/sensu-go/5.0/api/events.md
+++ b/content/sensu-go/5.0/api/events.md
@@ -570,7 +570,8 @@ curl -X PUT \
   "entity": {
     "entity_class": "proxy",
     "metadata": {
-      "name": "server1"
+      "name": "server1",
+      "namespace": "default"
     }
   },
   "check": {
--- a/content/sensu-go/5.1/api/events.md
+++ b/content/sensu-go/5.1/api/events.md
@@ -570,8 +570,7 @@ curl -X PUT \
   "entity": {
     "entity_class": "proxy",
     "metadata": {
-      "name": "server1",
-      "namespace": "default"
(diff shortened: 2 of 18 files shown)


hillaryfraley:
When I removed namespace from step 2 in the katacoda scenario, I got this error:
```
$ sensuctl configure -n \
> --username 'admin' \
> --password 'P@ssw0rd!' \
> --namespace default \
> --url 'http://localhost:8080'
$ sensuctl config view
=== Active Configuration
API URL:   http://localhost:8080
Namespace: default
Format:    tabular
Username:  admin
$ eval $(sensuctl env)
$ curl -sS -H 'Content-Type: application/json' \
> -H "Authorization: Bearer $SENSU_ACCESS_TOKEN" \
> -d '{
>   "entity": {
>     "entity_class": "proxy",
>     "metadata": {
>       "name": "server01"
>     }
>   },
>   "check": {
>     "metadata": {
>       "name": "server-health"
>     },
>     "output": "Hello, world!",
>     "status": 0,
>     "interval": 60
>   }
> }' \
> http://localhost:8080/api/core/v2/namespaces/default/events | jq .
{
  "message": "the namespace of the resource () does not match the namespace of the URI (default)",
  "code": 1
}
```

I think it's b/c we set the namespace as default in `sensuctl configure`. And I think you always set the namespace if you're using sensuctl, right? So maybe the namespace is required in the entity request body if you're using sensuctl?

hillaryfraley:
Hillary to scope back to 5.12 and later.

Need to look into Katacoda scenario issue.

hillaryfraley:
I removed versions prior to 5.12 from the PR.

cwjohnston:
@hillaryfraley I am seeing the same error, I've opened https://github.com/sensu/sensu-go/issues/3482 to track this.

hillaryfraley:
Replaced with https://github.com/sensu/sensu-docs/pull/2146 to resolve conflicts.

PR #2009, opened 2019-12-11
Fix keepalive-warning-timeout table

rgeniesse:
## Description
Removes an extra space that causes the `keepalive-warning-timeout` table to be broken.

## Motivation and Context
So the table renders properly.

## Review Instructions
Before:
Screen Shot 2019-12-11 at 11 34 32 AM

After:
Screen Shot 2019-12-11 at 11 34 17 AM


hillaryfraley (approved):
LGTM -- thank you for fixing this.

PR #2011, opened 2019-12-12
Add solaris package section to binary distro page

hillaryfraley:
## Description
Adds a Solaris section to the Binary-only distributions page in the 5.16 documentation.

## Motivation and Context
5.16 adds solaris/amd64 to the list of targets we build sensu-agent tarball releases for.

## Review Instructions
I copied the examples from the Linux section. Will also need to make sure the links I inserted for 36 and 37 are correct when the packages are actually released.


rgeniesse (approved):
LGTM. curl command works.

PR #2020, opened 2019-12-16
Add solaris info in Installation docs

hillaryfraley:
## Description
I left Solaris out in a couple places where it needed to be mentioned to close #2011 

## Motivation and Context
Additional changes based on #2011


rgeniesse (approved):
Nice catch!

PR #2030, opened 2019-12-17
Remove Go docs for 5.12

hillaryfraley:
## Description
Remove documentation for Sensu Go version 5.12 from docs.sensu.io/sensu-go

## Motivation and Context
Update docs site versions to conform to services agreement


cwjohnston (changes requested):
I believe this PR also needs to include an update to static.json, specifically [line 34](https://github.com/sensu/sensu-docs/blob/master/static.json#L34) where the regex should be changed to `"~ ^/sensu-go/5.1[0-2]/?((?<=\/).*)?$"`

cwjohnston:
Once static.json is included, we'll want to deploy this to Heroku staging for verification

hillaryfraley:
yes, you're correct! Thank you.

hillaryfraley:
Here's the branch deployed to staging: https://sensu-docs-stage.herokuapp.com/

I clicked through some links and tested that old versions and docs that don't exist in earlier versions will redirect to /latest.

I think it's worth burning one of our Moz reports to make sure. But I want to merge the guides and double-check a few link issues, then redeploy to staging so we get a clean Moz report.

cwjohnston (approved):

hillaryfraley:
This looks good to me. Only 15 4xx errors!
https://docs.google.com/spreadsheets/d/1cM-7vcEKQGImeKWaM6A-9oa4ojKAlrzTD4lhehCxn7c/edit#gid=196731968

PR #2043, opened 2019-12-20
Promote Core 1.9 to latest

hillaryfraley:
## Description
Promotes docs version Core 1.9 to latest

## Motivation and Context
Support Core 1.9.0 release

## Review Instructions
@cwjohnston I don't think I need to make any other changes to static.json (since we aren't archiving anything for Core), right?


cwjohnston (approved):
> I don't think I need to make any other changes to static.json (since we aren't archiving anything for Core), right?

I think that's right. LGTM 👍

Issue #1452, opened 2019-05-11
Epic: API examples


apaskulin:
Page: API docs

To do:
Test and add examples for the following endpoints

- [x] /assets POST
- [x] /assets/:asset (PUT)
- [x] /assets/:asset (DELETE)
- [x] /authproviders (GET)
- [x] /authproviders/:name (GET)
- [x] /authproviders/:name (PUT)
- [x] /clusterroles (POST)
- [x] /clusterroles/:clusterrole (PUT)
- [x] /entities (POST)
- [ ] /entities/:entity (POST)
- [x] /entities/:entity (PUT)
- [x] /events/:entity/:check (GET)
- [x] /etcd-replicators (POST)
- [x] /filters (POST)
- [x] /filters/:filter (PUT)
- [x] /handlers (POST
- [x] /handlers/:handler (PUT)
- [x] /hooks (POST)
- [x] /hooks/:hook (PUT)
- [x] /license (GET)
- [x] /license (PUT)
- [x] /license (DELETE)
- [x] /mutators (POST)
- [x] /mutators/:mutator (PUT)
- [x] /namespaces (POST)
- [x] /namespaces/:namespace (PUT)
- [x] /rolebindings (POST)
- [x] /rolebindings/:rolebinding (PUT)
- [x] /roles (POST)
- [x] /roles/:role (PUT)
- [x] /silenced (POST)
- [x] /silenced/:silenced (PUT)

hillaryfraley:
Next task: Update this list based on current state of docs.

hillaryfraley:
I updated the list to reflect the current state of the API docs as of 5.16.

PR #2055, opened 2019-12-30
Add POST example for etcd replicators in federation API doc

hillaryfraley:
Add POST /etcd-replicators example

Epic #1452


asachs01 (approved):
Initial pass LGTM 👍

PR #2056, opened 2019-12-30
Add POST and PUT examples in filters API doc

hillaryfraley:
Add POST /filters and PUT /filters/:filter examples.

Epic #1452

### DIFF
--- a/content/sensu-go/5.16/api/filters.md
+++ b/content/sensu-go/5.16/api/filters.md
@@ -27,20 +27,30 @@ The `/filters` API endpoint provides HTTP GET access to [event filter][1] data.
 The following example demonstrates a request to the `/filters` API endpoint, resulting in a JSON array that contains [event filter definitions][1].
 
 {{< highlight shell >}}
-curl http://127.0.0.1:8080/api/core/v2/namespaces/default/filters -H "Authorization: Bearer $TOKEN"
+curl -X GET http://127.0.0.1:8080/api/core/v2/namespaces/default/filters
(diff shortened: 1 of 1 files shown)


content/sensu-go/5.16/api/filters.md
```diff
@@ -105,12 +138,12 @@ payload         | {{< highlight shell >}}
   },
   "action": "deny",
   "expressions": [
-    "event.entity.metadata.namespace == 'production'"
+    "event.entity.metadata.namespace == '\''production'\''"
```
asachs01: This doesn't seem quite right to my eyes...are these meant to be escaped?

content/sensu-go/5.16/api/filters.md
```diff
@@ -183,7 +240,7 @@ payload         | {{< highlight shell >}}
...
```
asachs01: Also, this is stating that we're looking for a namespace called 'production', but looks like we're creating a 'development' filter...should we match this so that namespace == dev?

asachs01 (changes requested):

content/sensu-go/5.16/api/filters.md
hillaryfraley: I used Postman's copy button to copy the code, so it might be something Postman adds? In any case, I will remove it.

content/sensu-go/5.16/api/filters.md
hillaryfraley: yes, I think you're right, I'll fix that! Thank you.

Issue #846, opened 2018-10-30
Guide: Monitoring Sensu with Sensu


apaskulin:
Create a new guide at /guides/monitor-sensu based on https://docs.sensu.io/sensu-core/1.6/guides/monitor-the-monitor/

hillaryfraley:
PR to add requested guide: https://github.com/sensu/sensu-docs/pull/2071

PR #2071, opened 2020-01-03
Add Monitor Sensu with Sensu guide

hillaryfraley:
## Description
Add "Monitor Sensu with Sensu" guide

## Motivation and Context
Address https://github.com/sensu/sensu-docs/issues/846

## Review Instructions
I tried to adapt the Core guide (https://docs.sensu.io/sensu-core/latest/guides/monitor-the-monitor/) to Go. I will be happy to pair with someone to finish out the following content tasks:

- [ ] Add content for Sensu agent section
- [ ] Confirm all examples are correct and complete
- [ ] Add any missing content or clarifications and delete any irrelevant or incorrect information


content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
+  "api_version": "core/v2",
+  "metadata": {
+    "namespace": "default",
+    "name": "check_sensu_server_port"
+  },
+  "spec": {
+    "command": "check-http.rb -h remote-api-hostname -P 4567 -p /health?consumers=1 --response-code 204",
```
asachs01: Do we still have the concept of `consumers` in the Go API?

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
+  "api_version": "core/v2",
+  "metadata": {
+    "namespace": "default",
+    "name": "check_sensu_api_port"
+  },
+  "spec": {
+    "command": "check-ports.rb -H remote-sensu-server-hostname -p 4567",
```
asachs01: `:4567` isn't a valid port for Sensu Go--it's now `:8080`

```suggestion
    "command": "check-ports.rb -H remote-sensu-server-hostname -p 8080",
```

asachs01: We'd probably want to include etcd ports in this as well.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+    ],
+    "interval": 60,
+    "publish": true
+  }
+}
+{{< /highlight >}}
+
+## Monitor your Sensu agent instances
+
+**CONTENT NEEDED**
```
asachs01: * service checks
* port checks (if agent api is enabled)

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+    "subscriptions": [
+      "monitor_remote_sensu_api"
+    ],
+    "interval": 60,
+    "publish": true
+  }
+}
+{{< /highlight >}}
+
+## Monitor your Sensu dashboard instances
```
asachs01: I'd ditch this section since we don't decouple the dashboard from a backend any more.

asachs01 (changes requested):
I'll have some more, but this is my first pass.

content/sensu-go/5.16/guides/monitor-sensu.md
asachs01: I don't believe we do so this wouldn't be a valid check.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+version: "5.16"
+product: "Sensu Go"
+platformContent: False
+menu: 
+  sensu-go-5.16:
+    parent: guides
+---
+
+
+- [Monitor your Sensu backend instances](#monitor-your-sensu-backend-instances)
```
asachs01: Checks that could potentially comprise a fully-operational battlestation:
* Service checks for `sensu-backend`
* Port checks for the backend component ports
* Test queries against the api
* Prom exporter check against the /metrics endpoint

asachs01: Some additional checks here depending on if using embedded etcd or not. If using external etcd, there would be port/service(maybe) checks that would need to be relocated. 

Also, what about federation? We're in the practice of recommending a gateway be stood up--we'd need to figure out what would need to be monitored for a gateway that isn't processing events. Moreover, we might want a check that looks to see if a gateway somehow starts processing events, as that would indicate something is misconfigured

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+- [Monitor your Sensu agent instances](#monitor-your-sensu-agent-instances)
+- [Monitor your Sensu API instances](#monitor-your-sensu-api-instances)
+- [Monitor your Sensu dashboard instances](#monitor-your-sensu-dashboard-instances)
+
+This guide describes best practices and strategies for monitoring Sensu with Sensu. 
+It explains how to ensure your Sensu components are properly monitored, including your Sensu backend, agent, API, and dashboard instances.
+
+To completely monitor a Sensu stack (Sensu backend, agent, API, and dashboard), you will need to have at least one other independent Sensu instance.
+A single Sensu stack cannot monitor itself completely because if some components are down, Sensu cannot create events.
```
asachs01: While true, this statement reads to me as if it's referring to the old stack. Unless a user is using an external etcd cluster, there's not really `components` that fail individually. AFAIK if a backend is down, it's down.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+
+- [Monitor your Sensu backend instances](#monitor-your-sensu-backend-instances)
+- [Monitor your Sensu agent instances](#monitor-your-sensu-agent-instances)
+- [Monitor your Sensu API instances](#monitor-your-sensu-api-instances)
+- [Monitor your Sensu dashboard instances](#monitor-your-sensu-dashboard-instances)
+
+This guide describes best practices and strategies for monitoring Sensu with Sensu. 
+It explains how to ensure your Sensu components are properly monitored, including your Sensu backend, agent, API, and dashboard instances.
+
+To completely monitor a Sensu stack (Sensu backend, agent, API, and dashboard), you will need to have at least one other independent Sensu instance.
```
asachs01:
```suggestion
To completely monitor a Sensu stack (Sensu backend, agent, external etcd cluster (if used)), you will need to have at least one other independent Sensu instance.
```

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+
+_**NOTE**: This guide assumes you are not using Sensu clustering. The strategies in this guide are still useful for monitoring cluster members, but you should follow [Run a Sensu cluster][17] to effectively monitor clustered instances._
+
+## Monitor your Sensu backend instances
+
+The host running the `sensu-backend` should be monitored in two ways:
+
+* Locally by a `sensu-agent` process for operating system checks/metrics and Sensu services that are not part of the Sensu event system (like the dashboard)
+* Remotely from an entirely independent Sensu stack to monitor Sensu components that must be running for Sensu to create events. 
+
```
asachs01: Ok, so the only way that I can think of visualizing this without a wall of text would be a table:

|                                                     | service checks | port checks | prometheus collector check | api queries as checks |
|-----------------------------------------------------|----------------|-------------|----------------------------|-----------------------|
| Singular backend                                    |                |             |                            |                       |
| Clustered backends                                  |                |             |                            |                       |
| Distributed singular backend instances              |                |             |                            |                       |
| Distributed clustered backend instances             |                |             |                            |                       |
| Distributed singular backend instances (federated)  |                |             |                            |                       |
| Distributed clustered backend instances (federated) |                |             |                            |                       |

I'm only one ☕️ in , so this might be a bad way of doing this, but I'll defer to @sensu/reliability for some sanity checking here.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+- [Monitor your Sensu backend instances](#monitor-your-sensu-backend-instances)
+- [Monitor your Sensu agent instances](#monitor-your-sensu-agent-instances)
+- [Monitor your Sensu API instances](#monitor-your-sensu-api-instances)
+- [Monitor your Sensu dashboard instances](#monitor-your-sensu-dashboard-instances)
+
+This guide describes best practices and strategies for monitoring Sensu with Sensu. 
+It explains how to ensure your Sensu components are properly monitored, including your Sensu backend, agent, API, and dashboard instances.
+
+To completely monitor a Sensu stack (Sensu backend, agent, external etcd cluster (if used)), you will need to have at least one other independent Sensu instance.
```
hillaryfraley:
```suggestion
To completely monitor a Sensu stack [Sensu backend, agent, external etcd cluster (if used)], you will need to have at least one other independent Sensu instance.
```

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: I will remove it. I left a lot of stuff in here from the Classic doc b/c I didn't know what might and might not apply. Thank you!

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,163 @@
...
+- [Monitor your Sensu backend instances](#monitor-your-sensu-backend-instances)
+- [Monitor your Sensu agent instances](#monitor-your-sensu-agent-instances)
+- [Monitor your Sensu API instances](#monitor-your-sensu-api-instances)
+- [Monitor your Sensu dashboard instances](#monitor-your-sensu-dashboard-instances)
+
+This guide describes best practices and strategies for monitoring Sensu with Sensu. 
+It explains how to ensure your Sensu components are properly monitored, including your Sensu backend, agent, API, and dashboard instances.
+
+To completely monitor a Sensu stack [Sensu backend, agent, external etcd cluster (if used)], you will need to have at least one other independent Sensu instance.
+A single Sensu stack cannot monitor itself completely because if some components are down, Sensu cannot create events.
```
hillaryfraley:
```suggestion
```

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: Yeah -- would it work if I removed "consumers=1" here?

`"command": "check-http.rb -h remote-api-hostname -P 4567 -p /health --response-code 204",`

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,162 @@
...
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
+  "api_version": "core/v2",
+  "metadata": {
+    "namespace": "default",
+    "name": "check_sensu_api_port"
+  },
+  "spec": {
+    "command": "check-ports.rb -H remote-sensu-server-hostname -p 8080",
```
hillaryfraley: Aaron suggested using etcd port numbers here too.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Outside of the API port, you can't really monitor the agent as if the agent is down, a service check for it won't run. The keepalive is a built in monitor for the service being up and functional, at least partially (as network issues can cause a keepalive to go non-ok but the agent is fine).

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,162 @@
...
+
+- [Monitor your Sensu backend instances](#monitor-your-sensu-backend-instances)
+- [Monitor your Sensu agent instances](#monitor-your-sensu-agent-instances)
+- [Monitor your Sensu API instances](#monitor-your-sensu-api-instances)
+- [Monitor your Sensu dashboard instances](#monitor-your-sensu-dashboard-instances)
+
+This guide describes best practices and strategies for monitoring Sensu with Sensu. 
+It explains how to ensure your Sensu components are properly monitored, including your Sensu backend, agent, API, and dashboard instances.
+
+To completely monitor a Sensu stack [Sensu backend, agent, external etcd cluster (if used)], you will need to have at least one other independent Sensu instance.
```
rgeniesse: We may want to take out the use of the word stack here, as that was really about the high number of components needed to run SE Classic (SE, SE dashboard, rabbitmq, redis). Where now the requirement for most users is going to be just the Sensu Backend with the internal etcd.

We may want to eventually set up a section for the other possible setups (external etcd, postgres).

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,162 @@
...
+
+This guide requires Sensu plugins.
+For more information about using Sensu plugins, see [Install plugins with assets][16].
+
+_**NOTE**: This guide assumes you are not using Sensu clustering. The strategies in this guide are still useful for monitoring cluster members, but you should follow [Run a Sensu cluster][17] to effectively monitor clustered instances._
```
rgeniesse: The guide linked there doesn't tell you how to effectively monitor a Sensu Go instance that is clustered. There is a section called [here](https://docs.sensu.io/sensu-go/latest/guides/clustering/#manage-and-monitor-clusters-with-sensuctl), bit that heading is misleading; all those commands require a user to actively monitoring the cluster health, which this guide really isn't about.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: I think this is the best way to do a high level over view, then we could provide an example for each column or maybe have a suggested plugin column. However, I can see some overlap happening if we bring prometheus into this, although that's probably the best way to get etcd metrics out of Sensu Go.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: The port would also change from 4567 to 8080 for default instances. As is, it is a super basic check with just hitting `/health` as it would only tell you if the API is up. `/health` actually gives you more details like member health, but that would require an analysis of the response data to be useful via a program or jq.

We'd also want to try hitting port 3000 as well to monitor the dashboard portion and possible other API end points and etcd metrics (which might be covered above this).

rgeniesse: Actually it looks like we have an API section down below, we we may need to restructure this guide so that we don't confuse users where we are monitoring the API in two places depending on how we are monitoring `/health`.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: I'd say those should be in a the backend section then, as etcd isn't really the Sensu API, although I suppose it may not work if etcd is down.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: We still want to monitor port 3000 somewhere though and most likely from another Sensu Go instance to catch firewall issues. However, the need to check for a dashboard service is no longer needed.

rgeniesse (changes requested):
Check out in line comments.

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: Thank you -- I will update this and add placeholders for External etcd and PostgreSQL.

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: Thank you -- I removed the last clause and combined the rest.

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: I added a draft table, although I used different headings because otherwise the table columns aren't wide enough to accommodate the text. The content in the table cells is just placeholder and should be updated. If I misunderstood the purpose of this table, please let me know and I can fix it!

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: I added this content.

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: I updated the port numbers and removed "consumers=1" but I am not clear on what changes I need to make to this example so that it doesn't contradict the API example below or confuse readers. I will need more help with this point.

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: Would that be 2379 and 2380? I'm also not sure about -H vs. -h and -P vs. -P (the example under backend uses -h and -P)

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: I am not sure exactly how to incorporate this content. I did add the table you suggested -- maybe that is a good jumping-off point?

hillaryfraley:
@asachs01 and @rgeniesse I incorporated your comments, but there were a couple that I didn't know what to do with (and I added replies to explain those). My instinct is this still needs quite a bit more work -- wdyt?

rgeniesse:
@hillaryfraley thanks! Might be worth a pairing session or maybe an office hours topic (which I can be sure to be on) at this point to get all these comments on the same page. Most of this is just due to how Classic and Go are architecturally structured so different.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
+  "api_version": "core/v2",
+  "metadata": {
+    "namespace": "default",
+    "name": "check_sensu_server_port"
+  },
+  "spec": {
+    "command": "check-http.rb -h remote-api-hostname -P 8080 -P 3000 -p /health --response-code 204",
```
hillaryfraley:
```suggestion
    "command": "check-http.rb -h remote-api-hostname -P 8080 -p /health --response-code 204",
```

content/sensu-go/5.16/guides/monitor-sensu.md
hillaryfraley: Removed 3000 in command

hillaryfraley:
CREs to set up a time to review this doc, step through the different ports the backend listens on and which API points to query. There may be other checks (e.g., etcd-specific checks, cluster-specific checks, etc.) to add that users will want to query based on what they want to know. May need to separate overview table into subsections.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
+  "api_version": "core/v2",
+  "metadata": {
+    "namespace": "default",
+    "name": "check_sensu_server_port"
+  },
+  "spec": {
+    "command": "check-http.rb -h remote-api-hostname -P 8080 -p /health --response-code 204",
```
rgeniesse: Response code expected should be a 200 instead of a 204.

rgeniesse: Also add the `-w` flag to print the output of the check, which would give a user more context on the actual failure.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+Monitor the `sensu-backend` server process from an independent Sensu instance.
+To do this, call Sensu's [health API endpoint][6] and use the [check-http plugin][7] with the following check definition:
+
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
+  "api_version": "core/v2",
+  "metadata": {
+    "namespace": "default",
+    "name": "check_sensu_server_port"
```
rgeniesse:
```suggestion
    "name": "check_backend_health"
```

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+  },
+  "spec": {
+    "command": "check-http.rb -h remote-api-hostname -P 8080 -p /health --response-code 204",
+    "subscriptions": [
+      "monitor_remote_sensu_api"
+    ],
+    "interval": 60,
+    "publish": true
```
rgeniesse: Add `timeout` set to 10 seconds in the check definition.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+{{< /highlight >}}
+
+## Monitor your Sensu agent instances
+
+You can use API port checks to monitor your Sensu agent.
+Service checks are not effective for monitoring the Sensu agent because if the agent is down, a service check for it won't run.
+
+The [keepalive][11] is a built-in monitor for whether the Sensu agent service is running and functional, although network issues can cause a non-OK keepalive event even if the agent is functional.
+
+## Monitor your Sensu API instances
```
rgeniesse: This section can go away due to API health check covering the only check in this.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: This depends on whether we can monitor etcd cluster health through the `health` API endpoint.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Need to find out for the Dashboard portion if there are unauthenticated endpoint(s) to hit over HTTP to check for health.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+{{< highlight json >}}
+{
+  "type": "CheckConfig",
+  "api_version": "core/v2",
+  "metadata": {
+    "namespace": "default",
+    "name": "check_dashboard_port"
+  },
+  "spec": {
+    "command": "check-ports.rb -H dashboard-remote-hostname -p 3000",
```
rgeniesse: Won't need if we can get a http check going.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+    ],
+    "interval": 60,
+    "publish": true
+  }
+}
+{{< /highlight >}}
+
+## External etcd PLACEHOLDER
+
+Note from Richie: We may want to eventually set up a section for the other possible setups (external etcd, postgres).
```
rgeniesse: We should check with Engineering if there is any value in a backend monitoring its own etcd availability or if the health endpoint is enough from another cluster.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,150 @@
...
+
+## PostgreSQL PLACEHOLDER
+
+Note from Richie: We may want to eventually set up a section for the other possible setups (external etcd, postgres).
```
rgeniesse: Need to understand failure situations of postgres going down as it only processes events.
- Expected behavior of etcd if postgres goes away.
- Whether errors in the connection to postgres are surfaced through the health endpoint.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Agent websocket API check similar to backend `health` endpoint check. Need to check if there is one and if there isn't, ask for one through a FR against Sensu Go.

hillaryfraley:
[Notes from today's meeting](https://github.com/sensu/sensu-docs/files/4099697/discussion-notes-monitor-sensu-with-sensu.txt)

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Moving to action item in document.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Moving to action item in document.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Moving to action item in document.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Moving to action item in document.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Action item in document.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Action item in document.

asachs01:
FYI @sensu/reliability there's additional interest in this guide from other customers, as I had a call with them today where they're looking for this exact info and were wondering about what recommendations we have.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,149 @@
...
+- [Monitor your Sensu agent instances](#monitor-your-sensu-agent-instances)
+- [Monitor your Sensu dashboard instances](#monitor-your-sensu-dashboard-instances)
+- [External etcd PLACEHOLDER](#external-etcd-placeholder)
+- [PostgreSQL PLACEHOLDER](#postgresql-placeholder)
+
+This guide describes best practices and strategies for monitoring Sensu with Sensu.
+It explains how to ensure your Sensu components are properly monitored, including your Sensu backend, agent, API, and dashboard.
+
+To completely monitor Sensu (a Sensu backend with internal etcd and an agent), you will need at least one other independent Sensu instance.
+The second Sensu instance will ensure that you are notified when the primary is down and vise versa.
```
cwjohnston:
```suggestion
The second Sensu instance will ensure that you are notified when the primary is down and vice versa.
```

hillaryfraley:
Make sure that check definitions and conventions in the guide are aligned with the guidelines for the monitoring-checks repo the DAs are working on.
https://github.com/sensu-community/monitoring-checks

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,149 @@
...
+
+This guide requires Sensu plugins.
+For more information about using Sensu plugins, see [Install plugins with assets][10].
+
+_**NOTE**: This guide assumes you are not using Sensu clustering, although the strategies in this guide are still useful for monitoring cluster members._
```
cwjohnston:
```suggestion
_**NOTE**: This guide describes approaches for monitoring a single backend. The strategies in this guide are also useful for monitoring individual members of a backend cluster._
```

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Monitor your Sensu backend instances
+
+Monitor the host running the `sensu-backend` in two ways:
+
+* Locally by a `sensu-agent` process for operating system checks and metrics and Sensu services that are not part of the Sensu event system.
```
cwjohnston: > Sensu services that are not part of the Sensu event system.

I'm not sure what this means

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,149 @@
...
+### Monitor the Sensu backend locally
+
+Monitor the host that runs `sensu-backend` like any other node in your infrastructure.
+This includes checks and metrics for [CPU][1], [memory][2], [disk][3], and [networking][4].
+Find more plugins at [Bonsai, the Sensu asset index][5].
+
+### Monitor the Sensu backend remotely
+
+Monitor the `sensu-backend` from an independent Sensu instance. This will allow you to know if the Sensu event pipeline, API or dashboard is not working.
+To do this, call Sensu's [health API endpoint][6] and use the [check-http plugin][7] with the following check definition:
```
cwjohnston:
```suggestion
To do this, use the [check-http plugin][7] to query Sensu's [health API endpoint][6] with a check definition like this one:
```

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,149 @@
...
+    "subscriptions": [
+      "monitor_remote_dashboard_port"
+    ],
+    "interval": 60,
+    "publish": true
+  }
+}
+{{< /highlight >}}
+
+## Monitor your Sensu agent instances
```
cwjohnston: I think we can remove this section on monitoring agent state. Instead, I think we should add a note to the introduction indicating that agent status is out of scope for this guide, potentially linking to documentation about keepalives which are our best method for keeping tabs on agent state.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,149 @@
...
+You can use API port checks to monitor your Sensu agent.
+
+TODO: So the below statement is a bit tricky. While true for knowing if the agent is down, agents run on the majority of a users boxes, so they'll still most likely want to run normal checks (CPU, memory, disk, etc). But are those really "checking" Sensu? Or whatever that box is being used for?
+
+Service checks are not effective for monitoring the Sensu agent because if the agent is down, a service check for it won't run.
+The [keepalive][11] is a built-in monitor for whether the Sensu agent service is running and functional, although network issues can cause a non-OK keepalive event even if the agent has no other problems happening.
+
+## External etcd PLACEHOLDER
+
+TODO: See what makes sense here from a monitoring point of view, as since this is both external and a need for events to work for non-Postgres users, knowing the health of this will probably depend our secondary Sensu instance we talk about in this guide.
```
cwjohnston: I believe that we should be providing examples of the [sensu-prometheus-collector](https://github.com/sensu/sensu-prometheus-collector) for gathering metrics from sensu-backend (i.e. via backend /metrics API endpoint port 8080). This pattern works for collecting etcd metrics when using the backend's embedded etcd, as well as for collecting metrics from external etcd (in this case etcd /metrics API port 2379).

cwjohnston: the [check-etcd.rb](https://github.com/sensu-plugins/sensu-plugins-etcd/blob/1.1.0/bin/check-etcd.rb) plugin also looks like a good one for remotely monitoring etcd cluster health. 

When using embedded etcd, same cluster health details are available in response from /health API endpoint, but we would need to write a check-http example that inspects the relevant JSON key/value pairs to determine health.

content/sensu-go/5.16/guides/monitor-sensu.md
```diff
@@ -0,0 +1,149 @@
...
+
+## PostgreSQL PLACEHOLDER
+
+TODO: Postgres now is part of the health endpoint. Need to see what that looks like and what it looks like when it is not available.
```
cwjohnston: https://github.com/wrouesnel/postgres_exporter looks interesting for metrics but maybe more complex than we should try to take on right now.

If we want to use the /health endpoint for this, we need to solve for inspecting the postgres parts of that response body, similar to the above comment re: embedded etcd cluster health.

cwjohnston (changes requested):
Here are a few suggestions and comments based on first pass review.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: Hm, I guess in my mind I had thought about monitoring the API and Dashboard locally. But if we are doing those off machine, I think that's a bit of overkill. All it would let you know about is a firewall issue and you'd have to check for both checks since we don't have aggregates spanning Sensu instances. I'll remove that line.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: That sounds great to me. It's easy to try to do all the things with this. Keeping it small and adding to it later, even if this exact thing is never in there, is best.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: I've at least got /metrics checks working with my skunkworks environment, although I didn't test with an actual remote etcd environment yet. I've moved onto Postgres so we can have a full skeleton by tomorrow as I feel like the check-etcd.rb should be rather trivial to add.

content/sensu-go/5.16/guides/monitor-sensu.md
rgeniesse: So I have this working in theory with [check-http-json.rb](https://github.com/sensu-plugins/sensu-plugins-http/blob/master/bin/check-http-json.rb). However, the postgres items (and even the cluster items) are in a key where the value is an array of hashes, so I am not sure how to get at the part I care about. I verified this works with a key where the value is something more straight forwards link an int.

Health output on 5.19 when postgres is configured:

```
{
  "Alarms": null,
  "ClusterHealth": [
    {
      "MemberID": 9882886658148555000,
      "MemberIDHex": "8927110dc66458af",
      "Name": "default",
      "Err": "",
      "Healthy": true
    }
  ],
  "Header": {
    "cluster_id": 4255616304056077000,
    "member_id": 9882886658148555000,
    "raft_term": 2
  },
  "PostgresHealth": [
    {
      "Name": "postgres",
      "Active": true,
      "Healthy": true
    }
  ]
}
```

So my only guess here is that we need `--key PostgresHealth.something.Healthy --value true` to get this to work. I tried replacing something with `[]` but that didn't work.

Or if there are other JSON parsers out there, I can try those out if this is unsupported. The plugin does have a deep metrics option, just not seeing a flag or example even there to get that far down.

hillaryfraley:
@hillaryfraley move this to the 5.21 doc set

content/sensu-go/5.21/guides/monitor-sensu.md
```diff
@@ -0,0 +1,139 @@
...
+api_version: core/v2
+metadata:
+  name: sensu-prom-metrics
+spec:
+  command: sensu-prometheus-collector -exporter-url http://localhost:8080/metrics
+  output_metric_format: influxdb_line
+  output_metric_handlers:
+    - influxdb
+  subscriptions:
+    - monitor_remote_dashboard_port
```
asachs01, line 87: We should probably rename this subscription. It's not a port monitoring check, but really a metric check . So something like `site_beta_sensu_backend_metrics`

content/sensu-go/5.21/guides/monitor-sensu.md
```diff
@@ -0,0 +1,139 @@
...
+## Monitor your Sensu backend instances
+
+Monitor the host running the `sensu-backend` in two ways:
+
+* Locally by a `sensu-agent` process for operating system checks and metrics.
+* Remotely from an independent Sensu instance for Sensu components that must be running for Sensu to create events.
+
+### Monitor the Sensu backend locally
+
+Monitor the host that runs `sensu-backend` like any other node in your infrastructure.
```
asachs01, line 42: I'm wondering if we need to include service check in combination w/ port checks. Given the issues we've seen (an hopefully fixed) recently where the service was running but internal components weren't, this would make sense to implement in my mind.

However, this opens up a broader conversation about Sensu internals (e.g., the API, authorization component, etc) IMO.

content/sensu-go/5.21/guides/monitor-sensu.md
```diff
@@ -0,0 +1,139 @@
...
+---
+
+- [Monitor your Sensu backend instances](#monitor-your-sensu-backend-instances)
+- [Monitor your Sensu dashboard instances](#monitor-your-sensu-dashboard-instances)
+- [Monitor your Embedded etcd](#monitor-your-embedded-etcd)
+- [Monitor your External etcd](#monitor-your-external-etcd)
+- [PostgreSQL PLACEHOLDER](#postgresql-placeholder)
+
+This guide describes best practices and strategies for monitoring Sensu with Sensu.
+It explains how to ensure your Sensu components are properly monitored, including your Sensu backend, agent, API, and dashboard.
```
asachs01, line 21: But why monitor Sensu? Chatted w/ @cwjohnston and we need to have a narrative around why a customer would want to monitor Sensu in the first place.

content/sensu-go/5.21/guides/monitor-sensu.md
```diff
@@ -0,0 +1,139 @@
...
+  interval: 10
+  timeout: 5
+  runtime_assets:
+    - sensu-prometheus-collector
+{{< /highlight >}}
+
+## PostgreSQL PLACEHOLDER
+
+TODO: Postgres now is part of the health endpoint. Need to see what that looks like and what it looks like when it is not available.
+
```
asachs01, line 126: * Also add port/service check
* Are there any database queries that we need to add for monitoring?
* Would PGSQL have a prom exporter to use?

content/sensu-go/5.21/guides/monitor-sensu.md
```diff
@@ -0,0 +1,139 @@
...
+api_version: core/v2
+metadata:
+  name: sensu-prom-metrics
+spec:
+  command: sensu-prometheus-collector -exporter-url http://external-etcd:2379/metrics
+  output_metric_format: influxdb_line
+  output_metric_handlers:
+    - influxdb
+  subscriptions:
+    - monitor_remote_dashboard_port
```
asachs01, line 114: Change this to something else since this isn't monitoring the dashboard.

asachs01 (changes requested):
Still have a bit to figure out here.

hillaryfraley:
@asachs01 is this superseded by https://github.com/sensu/sensu-docs/pull/2568?

asachs01:
@hillaryfraley I believe that is the case.

hillaryfraley:
Closed in favor of https://github.com/sensu/sensu-docs/pull/2568

PR #2101, opened 2020-01-14
Remove extra colons from notes in entities API reference

rgeniesse:
## Description
Remove extra colons from notes in entities API reference

## Motivation and Context
Remove extra punctuation.


hillaryfraley (approved):
Thank you for catching this, Richie!

Issue #2118, opened 2020-01-17
Vault secrets management guide


nikkictl:
Write an interactive/e2e/step-by-step guide for how to configure a Vault provider and access secrets within checks, mutators and handlers. Resources required: `secrets/v1.Secret`, `secrets/v1.Provider` (type `VaultProvider`), `core/v2.CheckConfig`, `core/v2.Mutator`, `core/v2.Handler`. Secrets can only be passed to checks if mTLS is configured for that agent/backend.

Issue #2116, opened 2020-01-17
secrets/v1.Provider resource reference (types Env and VaultProvider)


nikkictl:
Add reference documentation for the `secrets/v1.Provider` (type `Env`) provider resource. The `Env` provider is a built-in secrets provider that can retrieve backend environment variables as secrets.

type Env
```
---
type: Env
api_version: secrets/v1
metadata:
  name: env
spec: {}
```

type VaultProvider (see https://github.com/sensu/sensu-docs/pull/2121 for more context on each VaultProvider field)
```
---
type: VaultProvider
api_version: secrets/v1
metadata:
  name: vault1
spec:
  client:
    address: https://vaultserver.example.com:8200
    token: b1g53cr3t!
    version: v1
    tls:
      ca_cert: "/etc/ssl/certs/vault_ca_cert.pem"
    max_retries: 2
    timeout: 20s
    rate_limiter:
      limit: 10
      burst: 100
```

Issue #2113, opened 2020-01-17
secrets/v1.Secret resource reference


nikkictl:
Add reference documentation for the `secrets/v1.Secret` resource.
- `ID` is the identifying key for the provider to retrieve the secret.
- `Provider` is the name of the provider with this secret.
```
type: Secret
api_version: secrets/v1
metadata:
  name: sensu-ansible-token
  namespace: default
spec:
  id: ANSIBLE_TOKEN
  provider: env
```

PR #2121, opened 2020-01-17
Add examples for VaultProvider and vault secrets

echlebek:
This is a partial document that adds examples for VaultProvider, and for secrets that use the VaultProvider.

### DIFF
--- a/content/sensu-go/5.17/api/secrets.md
+++ b/content/sensu-go/5.17/api/secrets.md
@@ -1,0 +1,73 @@
+#### Example {#vaultprovider}
+
+/secrets/v1/providers/:name (PUT) |
+-----------------------|-------------------------------------------
+description            | Create or update the vault secret provider
+example url            | http://hostname:8080/api/enterprise/secrets/v1/providers/vault1
+payload                | {{< highlight json >}}
+{
+  "type": "vault_provider",
+  "api_version": "secrets/v1",
+  "metadata": {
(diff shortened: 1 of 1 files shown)


hillaryfraley:
@echlebek it seems like this would eventually be merged into https://github.com/sensu/sensu-docs/pull/2122 -- is that correct?

nikkictl (commented):
I feel like the context you provide around the examples is very valuable. However I think that information might be more appropriate in reference docs or user guides, rather than api docs which just show each api endpoint. See https://github.com/sensu/sensu-docs/issues/2118, https://github.com/sensu/sensu-docs/issues/2116, and https://github.com/sensu/sensu-docs/issues/2113. @hillaryfraley wdyt?

hillaryfraley:
@nikkixdev @echlebek Is VaultProvider a separate endpoint from Secrets and Provider? I'm not quite clear on that.

nikkictl:
@hillaryfraley a VaultProvider is the same endpoint of a Provider. Similar to AuthProviders, you can use the same endpoints for LDAP and OIDC provider types. With secrets Providers, we have Env and Vault (see `Type` field).

hillaryfraley:
I think this content is covered in https://github.com/sensu/sensu-docs/pull/2126 and https://github.com/sensu/sensu-docs/pull/2125, and partly in https://github.com/sensu/sensu-docs/pull/2122.

PR #2132, opened 2020-01-23
Update UPN note in installation/auth

### LINKED ISSUE #2129: Update LDAP UPN security warning
### Description

5.17 release reintroduces support for LDAP UserPrincipalName (UPN). Using this approach to authenticate to LDAP has significant security implications which we need to call out more clearly.

### Category

This is an issue with:

### Affected pages

sensu-go/latest/installation/auth

### Expected behavior

When a setting has security implications, we will call out concrete examples of those implications.

### Current behavior

The note currently reads:

```
_WARNING: When using UPN authentication, users must re-authenticate to apply any changes made to group membership on the Active Directory server since their last authentication. To ensure group membership updates are reflected without re-authentication, specify a binding account or enable anonymous binding._
```

### Corrections needed

The note should explain that using UPN authentication creates situations where if employee is terminated or if they are removed from a group which has granted administrator permissions for their current session, that change will not be picked up by Sensu until the user logs out and tries to start a new session. Likewise, under UPN, users cannot be forced to log out of Sensu.
  
### Additional context

[LDAP Wiki article on UPN](https://ldapwiki.com/wiki/UserPrincipalName)

When UPN is in use, messages like these will be visible in sensu-backend logs:

* "UPN binding is enabled for server %s:%d, users group membership will not be updated" (where %s:%d are replaced with hostname and port for the LDAP server)
* "UPN binding for LDAP is currently enabled, claims will not be refreshed"

hillaryfraley:
## Description
Updates warning in `default_upn_domain` description to give an example of how group membership changes work under UPN, such as in the case of a terminated employee.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2129

## Review Question
I'm not sure whether we should include the context you listed in the issue -- it seems like the log info could be useful.

### DIFF
--- a/content/sensu-go/5.17/installation/auth.md
+++ b/content/sensu-go/5.17/installation/auth.md
@@ -971,7 +971,7 @@ example      | {{< highlight shell >}}
 
 | default_upn_domain |     |
 -------------|------
-description  | Enables UPN authentication when set. The default UPN suffix that will be appended to the username when a domain is not specified during login (for example, `user` becomes `user@defaultdomain.xyz`). _**WARNING**: When using UPN authentication, users must re-authenticate to apply any changes made to group membership on the AD server since their last authentication. To ensure group membership updates are reflected without re-authentication, specify a binding account or enable anonymous binding._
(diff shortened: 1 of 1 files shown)


cwjohnston (commented):
LGTM! @palourde I think you were involved in the latest discussion on the security implications of UPN, will you please give this change a look?

palourde (approved):
👌

PR #2136, opened 2020-01-24
Add notification override note and link in Enterprise InfluxDB guide

hillaryfraley:
## Description
Adds note that annotation description can be overridden by the notification message in the check result.

## Motivation and Context
Discussion during Retro mtg today.

## Review Instructions
The "latest" links didn't seem to be working for Core docs, so I used 1.9 here.


hillaryfraley:
I didn't see anything similar to this in our Go InfluxDB guide.

content/sensu-enterprise/2.6/integrations/influxdb.md
```diff
@@ -31,7 +31,7 @@ annotation columns  |
 `status`            | check status
 `client`            | client name
 `action`            | action of the event
-`description`       | output from the check
+`description`       | output from the check (can be overwritten by the [notification][6] specified in the check result)
```
cwjohnston: "can be" feels too imprecise.There's an order of precedence followed when generating event summaries:

1. check.notification
2. check.description
3. check.output

If check.notification is present it will be used, regardless of the presence of check.description or check.output
If check.description is present it will be used instead of check.output
As a required element of a check result, check.output will always be available as a fallback

This logic applies to multiple integrations in Sensu Enterprise; perhaps all of them.

cwjohnston (changes requested):
I looked into the logic we're trying to describe here and found that it applies to many if not all of the Enterprise integrations. Please see my inline comment and let's reconsider how best to approach this.

cwjohnston (approved):

PR #2144, opened 2020-01-28
Update secrets management guide

hillaryfraley:
## Description
Adds screenshots and clarification throughout the Secrets management guide.

## Motivation and Context
I merged the initial secrets management guide while a reviewer had in-progress comments.


hillaryfraley:
@cwjohnston you mentioned the `--dedup-key` flag...I am not sure where we would set that up in the guide. @nikkixdev can you clarify that? It's at line 237.

Edit:  never mind -- I just saw your reply to Cameron in #reliability. I'll add the --token flag.

nikkictl:
https://github.com/sensu/sensu-pagerduty-handler/blob/master/main.go#L32-L49

content/sensu-go/5.17/guides/secrets-management.md
```diff
@@ -219,7 +234,7 @@ metadata:
   name: pagerduty
 spec:
   type: pipe
-  command: pagerduty-handler --dedup-key $DEDUP_KEY
+  command: pagerduty-handler --token $DEDUP_KEY
   secrets:
   - name: DEDUP_KEY
```
cwjohnston:
```suggestion
  - name: PD_TOKEN
```

content/sensu-go/5.17/guides/secrets-management.md
```diff
@@ -219,7 +234,7 @@ metadata:
   name: pagerduty
 spec:
   type: pipe
-  command: pagerduty-handler --dedup-key $DEDUP_KEY
+  command: pagerduty-handler --token $DEDUP_KEY
```
cwjohnston:
```suggestion
  command: pagerduty-handler --token $PD_TOKEN
```

cwjohnston (changes requested):
I suggest adopting a different variable name for the handler config to avoid confusion.

cwjohnston:
Where possible I think we should use un-obfuscated screenshots -- i.e. no blurring -- to help folks have an idea of what the string might look like. 

With regard to Vault, the dev server root token etc. shouldn't be a concern to share -- I think it changes every time you restart the dev server, but I could be wrong. For PagerDuty we can generate a new integration key just for the screenshot and then delete it.

hillaryfraley:
No problem -- I can use inspect to change a few letters/numbers in the on-page html.

cwjohnston (approved):
Thanks for these improvements, looks much better to me!

hillaryfraley:
Thank you! I agree, it's better.

PR #2146, opened 2020-01-29
Remove /events PUT and namespace from required minimum example

### LINKED ISSUE #1621: Update events API docs

hillaryfraley:
## Description

- Removes namespace from required minimum example under https://docs.sensu.io/sensu-go/5.17/api/events/#eventsentitycheck-put-parameters.
- Removes /entity PUT section (Simon confirmed this is not a valid endpoint)

I tested this call by sending the required minimum example without a namespace included in the body of the request. The entity was created successfully without the namespace, so the namespace is not actually required.

I have confirmed that Events do not require a namespace using our Learn in 15 katacoda scenario (https://docs.sensu.io/sensu-go/latest/getting-started/tutorial/).

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/1621

## Review Instructions
Replaces https://github.com/sensu/sensu-docs/pull/1995
You can test by removing namespace from the event definitions in https://docs.sensu.io/sensu-go/latest/getting-started/tutorial/


content/sensu-go/5.14/api/events.md
```diff
@@ -690,8 +651,7 @@ curl -X PUT \
   "entity": {
     "entity_class": "proxy",
     "metadata": {
-      "name": "server1",
-      "namespace": "default"
```
cwjohnston, line 694: [Per the sensu-go changelog](https://github.com/sensu/sensu-go/blob/master/CHANGELOG.md#5170---2020-01-28), I believe the removal of namespace as a minimum required attribute should only apply to 5.17 and later.

cwjohnston (changes requested):
[Per the sensu-go changelog](https://github.com/sensu/sensu-go/blob/master/CHANGELOG.md#5170---2020-01-28), I believe the removal of namespace as a minimum required attribute should only apply to 5.17 and later.

Please update the PR so that the namespace attribute is only removed from the 5.17 version of the docs.

hillaryfraley:
@cwjohnston a couple longwinded questions about this:

We have this listed in the release notes for 5.12.0:
`Event and entity resources can now be created via the API without an explicit namespace. The system will refer to the namespace in the request URL.`
It's also in the 5.12.0 changelog (https://github.com/sensu/sensu-go/blob/master/CHANGELOG.md#5120---2019-08-22). I investigated...

I found the issue where you reported that POST /events wasn't working without a namespace for 5.16.1 (https://github.com/sensu/sensu-go/issues/3482). Simon's PR indicated that this always worked for PUT verbs and only needed to be fixed for POST (https://github.com/sensu/sensu-go/pull/3499). The minimum required example is under a PUT verb in the docs.

All this to ask:

- Should we keep this minimum required example (and maybe even correct it back to 5.12) but update the heading to include the PUT verb?
- Should we add a minimum required example that excludes namespace for POST as well, but only for the 5.17 doc?

cwjohnston (approved):
@hillaryfraley thanks for the clarifying questions. In my prior review I wasn't carefully considering POST vs. PUT and your comments have helped me better understand the nuance here.

> - Should we add a minimum required example that excludes namespace for POST as well, but only for the 5.17 doc?

I don't think we should do this. In most cases folks will want to use PUT, as it both creates or updates. Adding an example for POST, which only creates events but is otherwise identical in payload requirements, feels superfluous at this time.

> - Should we keep this minimum required example (and maybe even correct it back to 5.12) but update the heading to include the PUT verb?

If you want to add the PUT verb to the heading, that's fine with me, but I'm approving this as-is.

hillaryfraley:
Got it. I thought I remembered trying something related to this PR that worked! Thank you for double-checking.

Issue #1806, opened 2019-10-02
Refine offline docs support


hillaryfraley:
## This is an issue with:

- [ ] Bug (site functionality or styling)
- [ ] Errata (fix needed for doc content)
- [ ] New content (guide wanted)
- [ ] Update (Add missing or refresh existing content)
- [x] Enhancement (add new site functionality)

## Issues identified in initial offline docs PDF
- [x] ~PDF is not picking up bullet formatting or indention in lists and TOCs~ (fixed in https://github.com/sensu/sensu-docs/pull/2152)
- [x] ~Emojis are not printed in PDF~ https://github.com/sensu/sensu-docs/pull/1894
- [x] ~In the PDF bookmarks list, all doc page headings are nested under a "localhost" folder/heading~ Limitations of using this method with Acrobat are going to require manually renaming this bookmark after PDFs are generated
- [x] ~PDF bookmarks list includes "Sensu Docs" as part of each page title~ (fixed in https://github.com/sensu/sensu-docs/pull/2152)
- [x] ~PDF prints "Contents" after the heading~ (just before the TOC) on the first page of each docs article (fixed in https://github.com/sensu/sensu-docs/pull/2152)
- [ ] PDF ignores NOTE, WARNING, and PRO TIP formatting (except italics)

## Context
Refine results from initial PR, which allows us to generate offline copies of docs in PDF (#1779 )

See attached PDF, which includes comments noting instances of each issue listed above.
[Updated Review Copy - Sensu Go 5.13 - Sensu Docs.pdf](https://github.com/sensu/sensu-docs/files/3682899/Updated.Review.Copy.-.Sensu.Go.5.13.-.Sensu.Docs.pdf)

cwjohnston:
Additional todos: 

- [x] Play-test the PDF on e-reader device, especially TOC @cwjohnston 
- [x] Provide alternative images for emojis @hillaryfraley

cwjohnston:
I have tested the PDF on my Kindle via Amazon's email gateway. The TOC and other links I tried at random worked as expected. 🎉

hillaryfraley:
Here are image versions (mac) for each of the two emojis we use in the docs
![cross-mark_274c](https://user-images.githubusercontent.com/19190208/66421575-1d93a800-e9d6-11e9-8fcd-cfe7729ae1f5.png)
[...]

PR #2152, opened 2020-02-03
Fix offline docs formatting issues

rnevius:
## Description

Updates offline docs templates and styles to resolve the following issues:

- [x] PDF is not picking up bullet formatting or indention in lists and TOCs
- [x] PDF bookmarks list includes "Sensu Docs" as part of each page title
- [x] PDF prints "Contents" after the heading (just before the TOC) on the first page of each docs article

## Motivation and Context

Updates #1806

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -116,6 +116,32 @@ module.exports = function(grunt) {
       });
   });
 
+  grunt.registerTask("hugo-build-offline", function() {
+    const done = this.async();
+    const toml = require('toml');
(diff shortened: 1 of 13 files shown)


hillaryfraley (approved):
Thank you, @rnevius ! I created offline PDFs of both 5.17 and 5.16 and confirmed that all three elements of this PR are reflected in the PDFs.

hillaryfraley:
@cwjohnston I think we are good to merge this, but I'm leaving it open for now in case you want to take a look first. I'll merge it first thing tomorrow unless you do it first!

Issue #643, opened 2018-08-02
Guide Wanted: Using Check Aggregates


asachs01:
## This is an issue with:

- [ ] Bug (site functionality or styling)
- [ ] Errata (fix needed for doc content)
- [x] New content (guide wanted)
- [ ] Update (Add missing or refresh existing content)

## Expected Behavior
Guide describing use of check aggregates is present.

## Current Behavior
Guide isn't present.

## Possible Solution
Write the guide

## Context
"As a Sensu user, I want to know how to monitor the overall health of a service or application, not just an individual check."
"As a Sensu user, it'd be helpful to have a guide describing how I can monitor the overall health of a service or application."

hillaryfraley:
Add as a Discourse post for Sensu Go

hillaryfraley:
Aaron is testing some changes from Todd and then should be able to publish.

Aaron and Edgar to voltron on this issue.

asachs01:
Closed by #2162

PR #2162, opened 2020-02-10
Aggregates content proofing

asachs01:
Initial proof of aggregates guide before pushing to discourse.


cwjohnston:
Closes #643

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+weight: 350
+version: "5.16"
+product: "Sensu Go"
+platformContent: false
+menu:
+  sensu-go-5.16:
+    parent: guides
+---
+
+Sensu provides the capability to monitor groups of checks or entities via aggregates. In Sensu Go, this is done through the use of labels. In this tutorial, we'll walk through configuring both an entity aggregate and a check aggregate.
```
hillaryfraley:
```suggestion
Sensu allows you to monitor groups of checks or entities via aggregates. In Sensu Go, you construct aggregates using labels. This tutorial explains how to configure an entity aggregate and a check aggregate.
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+# Entity Aggregate
+
+Let's start with configuring an entity.
```
hillaryfraley:
```suggestion
To start, configure an entity.
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+product: "Sensu Go"
+platformContent: false
+menu:
+  sensu-go-5.16:
+    parent: guides
+---
+
+Sensu provides the capability to monitor groups of checks or entities via aggregates. In Sensu Go, this is done through the use of labels. In this tutorial, we'll walk through configuring both an entity aggregate and a check aggregate.
+
+# Entity Aggregate
```
hillaryfraley:
```suggestion
## Entity aggregate
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+Let's start with configuring an entity.
+
+The first thing we'll need to do is add a label to our agent.
```
hillaryfraley:
```suggestion
Add a label to your agent:
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  sensu-go-5.16:
+    parent: guides
+---
+
+Sensu provides the capability to monitor groups of checks or entities via aggregates. In Sensu Go, this is done through the use of labels. In this tutorial, we'll walk through configuring both an entity aggregate and a check aggregate.
+
+# Entity Aggregate
+
+Let's start with configuring an entity.
+
```
hillaryfraley, line 19:
```suggestion
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+##
+# agent overview
+##
+
+labels:
+  server_type: "webservers"
+
+```
+
+In this example, we've added the label `server_type` with the value `"webservers"`--it's useful to know if a number of agents on our webservers stop reporting in rather than just a single node. Let's add a similar label to a check.
```
hillaryfraley:
```suggestion
This example adds the label `server_type` with the value `"webservers"`. It's useful to know when a number of agents on your webservers stop reporting in rather than just a single node.

 Next, add a similar label to a check.
```

hillaryfraley: With regard to "It's useful to know when a number of agents on your webservers stop reporting in rather than just a single node.", the connection between this point and adding the label "webservers" doesn't seem obvious.

A connecting point might be something like:
"...the value `"webservers"`. In practice, you would add this label to all of your webserver agents so that you would know when a number of agents on your webservers stop reporting in rather than just a single node."

I don't know if that's what you meant here...it may need further clarification (or maybe I missed the point entirely).

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+# Check Aggregate
```
hillaryfraley:
```suggestion
## Check aggregate
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+The check we'll use for our aggregate is an http metric check for a ficticious app site
```
hillaryfraley:
```suggestion
This example check aggregate uses an http metric check for a fictitious app site: 
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  - slack
+  runtime_assets:
+  - sensu-plugins/sensu-plugins-http
+  - sensu/sensu-ruby-runtime
+  subscriptions:
+  - linux
+
+```
+
+While Sensu is capable of monitoring singular entities, it is also capable of monitoring groups of entities (and checks) through the use of aggregates. _______________ . Let's get started.
```
hillaryfraley: I don't think you need this paragraph because you already introduced the idea of using aggregates to monitor groups of entities.

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+===
```
hillaryfraley: Should the content below be incorporated into (or replace?) the content above?

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Entity Aggregates
```
hillaryfraley:
```suggestion
## Entity aggregates
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+In order for a check or entity to be considered as part of an aggregate it must have a label, or set of labels assigned to it. We'll start by setting up a fictional scenario in which we have 20 webservers serving a number of applications. In this scenario, we might not care if a singular server stops responding, but we might care if say, 15 of the 20 stop responding.
```
hillaryfraley:
```suggestion
To include a check or entity in an aggregate, you must assign a label or set of labels to it. For this example, imagine you have 20 webservers serving a number of applications. In this scenario, you might not care if a single webserver stops responding, but you would care if 15 of the 20 webservers all stop responding. 
```

hillaryfraley: It would be good to add a bridge here -- something to connect the scenario to the entity you're setting up below. Something like this perhaps?
"Start by identifying which 20 entities belong to your webserver aggregate."

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+While Sensu is capable of monitoring singular entities, it is also capable of monitoring groups of entities (and checks) through the use of aggregates. _______________ . Let's get started.
+
+===
+
+## Entity Aggregates
+
+In order for a check or entity to be considered as part of an aggregate it must have a label, or set of labels assigned to it. We'll start by setting up a fictional scenario in which we have 20 webservers serving a number of applications. In this scenario, we might not care if a singular server stops responding, but we might care if say, 15 of the 20 stop responding. 
+
+To make an entity part of an aggregate, we would need to add a label in our `/etc/sensu/agent.yml`. For example:
```
hillaryfraley:
```suggestion
Add a label in your `/etc/sensu/agent.yml`:
```

hillaryfraley: This may be confusing because it doesn't explain why we're only adding a label for one agent and not 20.

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+##
+name: "webserver01.example.com"
+namespace: "default"
+subscriptions:
+  - webservers
+labels:
+  server_role: "webserver"
+```
+
+After adding the label, we'll make sure to restart our agent to pick up the change in configuration:
```
hillaryfraley:
```suggestion
After you add the label, restart your agent to pick up the change in configuration:
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+`systemctl restart sensu-agent`
+
+We'll need to set up a check that will look at events with the label we've assigned to the entity and ensure that these events are in an OK status. Let's look at an example check:
```
hillaryfraley:
```suggestion
Next, configure a check that will identify events with the label you assigned to the entity and ensure that these events are in an OK status. Here's an example:
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  - backend
+  publish: true
+  interval: 30
+  handlers:
+  - slack
+  - pagerduty
+  - email
+```
+
+The check command uses a username/password combination to access the API and then matches events with the label "server_role: webserver". In this case, the check will create an event if 75 percent of the aggregate events are in a `warning` state and will also create an event if 50 percent of the aggregate events are in a critical state.
```
hillaryfraley:
```suggestion
The check command uses a username/password combination to access the API and matches events with the label "server_role: webserver". The check will create an event if 75% of the aggregate events are in a `warning` state and if 50% of the aggregate events are in a critical state.
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+labels:
+  server_role: "webserver"
+```
+
+After adding the label, we'll make sure to restart our agent to pick up the change in configuration:
+
+`systemctl restart sensu-agent`
+
+We'll need to set up a check that will look at events with the label we've assigned to the entity and ensure that these events are in an OK status. Let's look at an example check:
+
```
hillaryfraley, line 46: Could it be confusing that we have not explained how events are part of this? We have explained that you need to add a label to an entity, but we haven't explained something like "The events this entity produces will be associated with the `webserver` label, which will allow us to watch for those events with a check."

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  handlers:
+  - slack
+  - pagerduty
+  - email
+```
+
+The check command uses a username/password combination to access the API and then matches events with the label "server_role: webserver". In this case, the check will create an event if 75 percent of the aggregate events are in a `warning` state and will also create an event if 50 percent of the aggregate events are in a critical state.
+
+## Check Aggregates
+
```
hillaryfraley, line 71: Maybe we could be clearer from the start that this guide explains two separate ways to construct and use aggregates, entity or check (if that is actually what we are explaining -- I may be confused about that).

If we are explaining two options, I also think we should add some basic info about why/when to use one or the other.

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+In addition to having entities comprise aggregates, checks can also comprise aggregates. Continuing with our scenario, let's suppose that our webservers are serving various applications on different ports: 80, 8080, 9000. A standard check grouping might look like:
```
hillaryfraley:
```suggestion
Checks can also comprise aggregates. To continue the scenario, suppose that your webservers are serving various applications on different ports: 80, 8080, and 9000. A standard check grouping might look like this:
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  low_flap_threshold: 0
+  publish: true
+  runtime_assets:
+  - sensu-plugins/sensu-plugins-http
+  - sensu/sensu-ruby-runtime
+  subscriptions:
+  - linux
+```
+
+So we have three separate checks that are monitoring our web application. However, if we wanted to view the webapp's health these three checks don't do the best job providing that insight. The checks are isolated from each other and each check alerts individually.
```
hillaryfraley:
```suggestion
Three separate checks are monitoring your web application. However, if you want view your webapp's health, these three checks don't do the best job of providing that insight. These checks are isolated from each other, and each check alerts individually. 
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  runtime_assets:
+  - sensu-plugins/sensu-plugins-http
+  - sensu/sensu-ruby-runtime
+  subscriptions:
+  - linux
+```
+
+So we have three separate checks that are monitoring our web application. However, if we wanted to view the webapp's health these three checks don't do the best job providing that insight. The checks are isolated from each other and each check alerts individually. 
+
+Instead, it makes more sense to have this group of checks be part of an aggregate--we might not care if a check on an individual host fails, but we certainly care if a large percentage of the checks are in a warning or critical state across a number of hosts.We can do this by adding a label to the checks:
```
hillaryfraley:
```suggestion
Instead, it makes more sense to configure this group of checks as an aggregate because you might not care if a check on an individual host fails, but you will certainly care if a large percentage of the checks are in a warning or critical state across a number of hosts.

To turn these checks into an aggregate, add a label to each of them:
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  low_flap_threshold: 0
+  publish: true
+  runtime_assets:
+  - sensu-plugins/sensu-plugins-http
+  - sensu/sensu-ruby-runtime
+  subscriptions:
+  - linux
+```
+
+Each check now has a label that we can use as part of an aggregate that gives us more visibility into the health of our webapp. You'll also notice that we removed handlers from the check--if we want to alert on an aggregate, we're better served having the aggregate handled versus having each individual check handled.
```
hillaryfraley:
```suggestion
You can use the labels as part of an aggregate that gives you more visibility into the health of our webapp. You also removed handlers from the check. If you want to alert on an aggregate, it's better to handle the aggregate instead of handling each individual check.
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+So to check these services as part of a combined aggregate, we'll use a check that looks like this:
```
hillaryfraley:
```suggestion
Now, to check these services as part of a combined aggregate, use a check like this:
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+  subscriptions:
+  - backend
+  publish: true
+  interval: 30
+  handlers:
+  - slack
+  - pagerduty
+  - email
+```
+Now that we've got our aggregate in place, let's take a look at what this might look like in the Sensu UI:
```
hillaryfraley:
```suggestion

Congratulations! Your aggregate is in place. Here's how it might look in the Sensu web UI:
```

content/sensu-go/5.16/guides/using-check-aggregates.md
```diff
@@ -0,0 +1,285 @@
...
+
+![TO DO INSERT DASHBOARD PIC HERE][]
```
hillaryfraley, line 227: Are you planning to add a similar example for the entity aggregate?

asachs01:
Content up in https://discourse.sensu.io/t/tutorial-using-check-aggregates-in-sensu-go/1748

PR #2174, opened 2020-02-13
Add shortcode and css for note, warning, pro tip, and important callouts

### LINKED ISSUE #2158: Static content should not be stylized via frontend JavaScript
### Description

Most documentation content is stylized by Hugo while rendering markdown as HTML, however 
some static content, i.e. `NOTE`, `PROTIP`, `IMPORTANT` and `WARNING` "callouts", are rendered using JavaScript instead.

### Category

This is an issue with:

### Affected pages

Any page with `NOTE`, `PROTIP` or `WARNING` "callouts"

### Expected behavior

Static content is stylized by the static site build process, without dependency on JavaScript.

### Current behavior

As described in [this comment](https://github.com/sensu/sensu-docs/issues/1806#issuecomment-581578080), the special formatting applied to content inside a `_NOTE: _` or similar "callout" is being generated by Javascript instead of being rendered in the HTML generated by Hugo.

### Changes needed

As I have modeled this, we will be best served by implementing this in phases:

- [ ] Implement new formatting template tags, potentially one for each callout
- [ ] Update the latest version of the documentation to use these new tags

Once we've archived the last version using the old formatting:

- [ ] Remove the JavaScript styling implementation

### Additional context

Rendering `NOTE`, `PROTIP` or `WARNING` "callouts" using JavaScript prevents the content from being rendered properly in offline layout, i.e. when we generate a PDF.

* Browser(s) and version(s): 
* OS:

hillaryfraley:
## Description
- Adds a shortcode file: sensu-docs/layouts/shortcodes/notice.html
- Adds `notice` classes to style sheet: sensu-docs/static/stylesheets/scss/_alert.scss
- Includes four examples (one for each callout) in /sensu-docs/content/sensu-go/5.17/getting-started/learn-sensu/ (see screenshot and explanation below)

Uses [Hugo Learn theme's notice.html shortcode](https://github.com/matcornic/hugo-theme-learn/blob/15f890999f768fd31307a53e25ccb857f9e559c3/layouts/shortcodes/notice.html), with alterations to the [theme css](https://github.com/matcornic/hugo-theme-learn/blob/master/static/css/theme.css) to fit our formatting.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2158

## Review Instructions
I created placeholder callouts to demonstrate all four styles implemented with the shortcode. If you run the site locally, you can see them at http://localhost:1313/sensu-go/5.17/getting-started/learn-sensu/ (they are the first four callouts in the document).

Currently, the callouts using shortcodes look like this:
notes-shortcode-example

You can see that the markdown links in the callout text are not rendering. Based on [this part of the Hugo docs](https://gohugo.io/content-management/shortcodes/#shortcodes-with-markdown), I thought the `{{ $_hugo_config := `{ "version": 1 }` }}` at the top of the shortcode file should handle it, but it doesn't. And besides that, it sounds like we should be using the newer `{{< >}}` instead of the older `{{% %}}`.

I also ran through [this thread](https://discourse.gohugo.io/t/misunderstanding-of-shortcode-syntax-in-0-55/18538/4), but I'm not sure I understand how to implement markdownify (or if that would solve this problem). I tried using `{{ .Inner | markdownify }}` in the shortcode, but it didn't work.

@rnevius, this is not a high priority, but I think I have gotten as far as I can with it and should turn it over to you at this point.


layouts/shortcodes/notice.html
```diff
@@ -0,0 +1,2 @@
+{{ $_hugo_config := `{ "version": 1 }` }}
```
cwjohnston, line 1:
```suggestion
```

layouts/shortcodes/notice.html
```diff
@@ -0,0 +1,2 @@
...
+<div class="notices {{ .Get 0 }}" {{ if len .Params | eq 2 }} id="{{ .Get 1 }}" {{ end }}>{{ .Inner }}</div>
```
cwjohnston:
```suggestion
<div class="notices {{ .Get 0 }}" {{ if len .Params | eq 2 }} id="{{ .Get 1 }}" {{ end }}>{{ .Inner | markdownify }}</div>
```

cwjohnston (changes requested):
I think changing `{{ .Inner }}` to `{{ .Inner | markdownify }}` will do the trick, and with that change we won't need the `$_hugo_config` either.

I gleaned this idea from https://discourse.gohugo.io/t/misunderstanding-of-shortcode-syntax-in-0-55/18538

hillaryfraley:
When I try it with the `{{ $_hugo_config := `{ "version": 1 }` }}` line removed and ` {{ .Inner | markdownify }}`, here is what I see:
with-markdownify

rnevius:
Looks good to me. ~I removed the now-superfluous JS~ EDIT: #2158 states that we should hold off on removing this JS for now, so I've reverted commit 72fecca

rnevius:
The shortcode template has been copied to `offline/` layouts

hillaryfraley:
@rnevius thank you. I think my initial commit message probably wasn't very clear. I have added the notice.html shortcode, which works for everything except the markdown links (see the image in my intial commit message).

Using `{{ .Inner | markdownify }}` in the shortcode file does not work. See the image in [my reply to Cameron](https://github.com/sensu/sensu-docs/pull/2174#issuecomment-588291763) above to see what happens when I try that solution.

I'm not sure what I'm doing incorrectly here. Do you know how to make the markdown links work inside the shortcode-formatted text?

rnevius:
@hillaryfraley I think we're on the same page...The markdown is rendering correctly for me when I checkout the most recent changes in this branch. Can you try opening the page in a private browser instance, or try running the server with the `--ignoreCache` flag?

hillaryfraley:
@rnevius thank you. I'm still seeing the same issues.

I redeployed the site to staging from this branch, and here's what I see in an incognito window (the page with the test messages is https://sensu-docs-stage.herokuapp.com/sensu-go/latest/getting-started/learn-sensu/):
2020-02-24_07-56-26-incognito

I also tried it locally with the --ignoreCache flag, and I see the same thing.

Then I tried it again locally with the flag using Cameron's suggested changes. Here is what I see when I do that:
2020-02-24_07-55-36-local

rnevius:
@hillaryfraley apologies, I do see the same issue. I [did some digging](https://discourse.gohugo.io/t/markdown-reference-links-in-shortcodes/5770/2) and it looks like the problem is that:

> The shortcodes and the surrounding content share no common (Blackfriday) context

In plain English, this means that shortcodes only have access to the markdown context within the shortcode block. Because we're using reference-style markdown links (e.g. `[some text][1]` where `[1]` is defined at the end of the document), Blackfriday does not have access to the `[1]` definition (as it exists outside of the shortcode block). The issue persists with the new Goldmark markdown rendering engine as well, in my local tests.

A workaround would be to leave the shortcode template as-is and change links within these shortcode blocks to the inline format: `[first time setup reference](../../sensuctl/reference/#first-time-setup)`.

hillaryfraley:
@rnevius thank you for getting to the bottom of it! That does indeed work:
2020-02-25_11-57-09-fixed

@cwjohnston I would like to go ahead with this change throughout the 5.18 docs (after the 5.18 release). Wdyt?

- [x] Replace formatting for all note, warning, tip, and important callouts with shortcodes
- [x] Replace all reference links with inline links in all shortcode-formatted note/warning/tip/important callouts

hillaryfraley:
@rnevius thank you for getting to the bottom of it! That does indeed fix the problem!
2020-02-25_11-57-09-fixed

@cwjohnston I would like to go ahead with this change throughout the 5.18 docs (after the 5.18 release). Wdyt?

- [ ] Replace formatting for all note, warning, tip, and important callouts with shortcodes
- [ ] Replace all reference links with inline links in all shortcode-formatted note/warning/tip/important callouts

hillaryfraley:
I deployed this branch to staging (https://sensu-docs-stage.herokuapp.com/) and clicked around to check that all the callouts look good in 5.18

cwjohnston:
@hillaryfraley sorry I missed your ping on this but it looks good to me. Thanks all!

PR #2175, opened 2020-02-14
Correct pagination overview and details in namespaces and silenced API doc

### LINKED ISSUE #2171: Pagination is only supported for core/v2 resources
### Description
Need to audit the API docs to make sure we only list pagination support for https://github.com/sensu/sensu-go/tree/master/api/core/v2 resources.

### Category

This is an issue with:

### Affected pages
Audit all pages in API docs section and update as needed.

### Expected behavior
Docs only list pagination support where it is actually available.

### Current behavior
This may already be correct in the docs, but we should audit to make sure.

### Corrections needed
Corrections pending audit results.

hillaryfraley:
## Description
- Replaces the namespaces pagination examples in the API overview with users examples so that they will work (namespaces examples will not work with `continue`)
- Updates the API overview explanation for `continue` param to explain that this is in the request **header**
- Updates the API overview pagination explanation to clarify that it is available for most core/v2 API endpoints that return an array
- Adds subheadings in pagination section of API overview for the two query parameters
- In namespaces API doc, updates pagination note in the GET /namespaces table to explain that the endpoint only works with the limit query param
- In silencing API doc, updates pagination note in the GET /silenced table to explain that the endpoint does not work with pagination

## Motivation and Context
Closes #2171


cwjohnston (approved):
Looks great to me! 🎉

PR #898, opened 2018-11-14
Update documentation for javascript filtering.

echlebek:
The filtering engine in sensu-go has been replaced with Javascript. This PR updates the documentation accordingly.

The PR is here and is expected to be included in Beta 8: https://github.com/sensu/sensu-go/pull/2336

Issue #1750, opened 2019-09-12
Elaborate on being able to use golang regex for asset filters and filters


asachs01:
## This is an issue with:

- [ ] Bug (site functionality or styling)
- [ ] Errata (fix needed for doc content)
- [ ] New content (guide wanted)
- [x] Update (Add missing or refresh existing content)
- [ ] Enhancement (add new site functionality)

## Expected Behavior

Have a section in our filters ref that calls out being able to use golang regex a la https://github.com/google/re2/wiki/Syntax

## Current Behavior

This currently doesn't exist

## Possible Solution

Write it ™️ (beautifully)

cwjohnston:
Next steps
- [x] Set up a test environment to prove out this concept (using golang regex in asset filters)
- [ ] Determine the right place for this information
- [ ] Add content in the right place

cwjohnston:
## Expected behavior

An EventFilter which specifies one or more expressions using `match(regex)`, where `regex` is a Go RE2-compatible expression, should function as a filter.

## Context

To test, I set up this Handler and EventFilter:

```
---
type: Handler
api_version: core/v2
metadata:
  name: debug
spec:
  type: pipe
  command: echo
  filters:
  - allow_foo
---
type: EventFilter
api_version: core/v2
metadata:
  name: allow_foo
  namespace: default
spec:
  action: allow
  expressions:
    - event.check.labels.filter_test.match(/foo.*/)
```

Now, post an event with a matching label to the Sensu API:

```
curl -sS -H 'Content-Type: application/json' -H "Authorization: Bearer $SENSU_ACCESS_TOKEN" -d '{
  "entity": {
    "entity_class": "proxy",
    "metadata": {
      "name": "server01",
      "namespace": "default"
    }
  },
  "check": {
    "metadata": {
      "name": "server-health",
      "labels": {
        "filter_test": "seafood"
      }
    },
    "output": "Hello, world!",
    "status": 0,
    "interval": 60,
    "handlers": ["debug"]
  }
}' http://localhost:8080/api/core/v2/namespaces/default/events
```

Now note that the event is allowed by the backend:

```
{"check_name":"server-health","check_namespace":"default","component":"pipelined","entity_name":"server01","entity_namespace":"default","hooks":null,"level":"debug","msg":"received event","silenced":[],"time":"2020-01-29T23:39:56Z","timestamp":1580341196}
[...]

PR #2182, opened 2020-02-14
Improve Sensu query expressions doc and add label/annotation examples

### LINKED ISSUE #1061: Docs Feedback: Using annotations in query expressions
Link to page: https://docs.sensu.io/sensu-go/5.1/reference/sensu-query-expressions/

### Feedback
No example on how to use check annotations in SQE.  I've looked in the following areas:
https://docs.sensu.io/sensu-go/5.1/reference/filters/
https://docs.sensu.io/sensu-go/5.1/reference/checks/
and the above link for sensu-query-expressions.

hillaryfraley:
## Description
I want to add more explanation for SQEs and include useful examples for operators. I think I need to talk this through with someone who understands it better than me. Draft for now.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/1061

## Review Instructions
I have some questions about this doc and SQEs in general. I searched the sensu-go and sensu-docs repos for some history (https://github.com/sensu/sensu-go/issues/2328, https://github.com/sensu/sensu-go/pull/2336, and https://github.com/sensu/sensu-docs/pull/898) but didn't find much to go on.

- How do you know which attributes you can use in an SQE? Do we have a list somewhere? Seems like some of this info is scattered among various docs, like [filters](https://docs.sensu.io/sensu-go/latest/reference/filters/#build-event-filter-expressions). Is this the complete list of attributes you can use? If so, I should update the doc to explain that you can only use Sensu query expressions for specific event filter attributes.
  - Should the references in other docs be updated to use the terminology "Sensu query expressions"?
- How do you tell the SQE which event/entity/filter/etc. you want it to evaluate? Do users need to install the Otto JS VM?
- Why do our examples use different highlighting? The custom functions are highlighted for Go, examples are highlighted for JS or text.
- How are SQEs different from API response filtering? Is it just that you don't have to call the API to use them?
- Under "Specification" we say `an error is logged` -- where is the error logged? And is it logged by Sensu, or elsewhere?
- I need a better understanding of the differences and reasons for using labels vs. annotations. I added a couple examples to this doc, but I don't know if they're accurate or represent a useful example for operators.


content/sensu-go/5.17/reference/sensu-query-expressions.md
```diff
@@ -12,12 +12,15 @@ menu:
...
-- [Sensu query expression examples](#sensu-query-expression-examples)
+- [Specification](#specification)
+  - [Custom functions](#custom-functions)
+- [Examples](#examples)
 
-Sensu query expressions are based on [JavaScript][1] expressions.
-They provide additional functionality for Sensu usage, like nested parameters and custom functions, so Sensu resources can be directly evaluated.
-Sensu query expressions should always return `true` or `false`.
+Sensu query expressions are JavaScript-based expressions that provide additional functionality for using Sensu, like nested parameters and custom functions.
+You can use Sensu query expressions to directly evaluate Sensu resources.
```
cwjohnston:
```suggestion
```

Even in the original content this statement doesn't convey meaning to me. Perhaps aspirational, that one could write JavaScript library to interact with the Sensu API. Since we don't have meaningful examples for this right now, I feel like we should strike it to avoid confusion.

cwjohnston:
> * How do you know which attributes you can use in an SQE? Do we have a list somewhere? 

This is a bit tricky. The expressions are evaluated by the Otto VM as JavaScript programs. The Otto VM is [mostly but not entirely compatible with the ECMAScript 5 specification](https://github.com/robertkrimen/otto#caveat-emptor).

> Seems like some of this info is scattered among various docs, like [filters](https://docs.sensu.io/sensu-go/latest/reference/filters/#build-event-filter-expressions). Is this the complete list of attributes you can use? If so, I should update the doc to explain that you can only use Sensu query expressions for specific event filter attributes.

I think the attributes listed in the event filter reference are fairly representative of what is possible. Any time we add attributes to [those in the Event spec](https://docs.sensu.io/sensu-go/latest/reference/events/#spec-attributes), I expect those attributes to be available to event filter expressions.

Naturally we want our list of event attributes to be as complete as possible, but there's a large number of JavaScript functions or methods that can be used to take action on the values of event attributes, and without knowledge of the JavaScript/ECMA5 standard library, it's not always obvious when one is calling a method on an object vs accessing an attribute value.

>   * Should the references in other docs be updated to use the terminology "Sensu query expressions"?

I suspect the name "Sensu Query Expressions" is intended to encapsulate and convey some of the above nuance, but I would like us to better explain what it means on this page before we consider expanding it to other pages.

> * How do you tell the SQE which event/entity/filter/etc. you want it to evaluate? Do users need to install the Otto JS VM?

I believe SQEs are only evaluated in the context of EventFilters, so they always receive a single event as `event`, which event will also contain some information about the entity which collected the check result. etc. Accessing information about other events, entities, etc. within an SQE will require access to the API. We don't currently have workable examples of API access from within an SQE context.

> * Why do our examples use different highlighting? The custom functions are highlighted for Go, examples are highlighted for JS or text.

Not sure but it sounds like we should normalize it

> * How are SQEs different from API response filtering? Is it just that you don't have to call the API to use them?

SQEs are in some ways limited to evaluating events in the event pipeline. API response filters are applicable to a wider variety of resources but generally don't influence the processing of events in the pipeline.

SQEs are defined in EventFilters, so they are active in the context of the event pipeline, determining whether or not a given event should be passed to the handler. 

API response filtering is not part of the event pipeline, but instead allows API clients to specify constraints which limit the resources returned by a given API request. 

> * Under "Specification" we say `an error is logged` -- where is the error logged? And is it logged by Sensu, or elsewhere?

I believe in all cases messages like these would be in the sensu-backend log.

cwjohnston:
> * I need a better understanding of the differences and reasons for using labels vs. annotations. I added a couple examples to this doc, but I don't know if they're accurate or represent a useful example for operators.

I will re-read with an eye for this and see if I can make any suggestions. I believe this would also be a good place to fold in some examples making use of [golang regular expressions](https://github.com/sensu/sensu-docs/issues/1750).

cwjohnston (changes requested):

hillaryfraley:
@cwjohnston thank you. I updated the doc to incorporate your answers to my questions, with the exception of the "available attributes" (which I need to think about a bit more).

cwjohnston (approved):

PR #2185, opened 2020-02-18
Update federation guide diagram to correct etcd port number

cwjohnston:
## Description

Fix an error in the existing federation diagram related to the etcd port number.

## Motivation and Context

Customer reported confusion due to differences between the port number for EtcdReplicators indicated in the diagram (2780) and the port used in the guide (2379)

## Review Instructions

Requires visual review outside github. etcd port number in the diagram legend/key should be the only change


Blue0ctober (approved):
👍

Blue0ctober (approved):

PR #2275, opened 2020-03-15
[WIP] Add Ansible examples to Sensu Go docs

tadeboro:
## Description

This PR introduces Ansible examples into the Sensu Go docs alongside YAML and JSON ones.

## Motivation and Context

Users that are trying to automate their Sensu Go deployments using Ansible from the get-go need to consult two different documents: one with the Sensu Go information and one with the Ansible collection information. Placing sample Ansible tasks alongside the YAML and JSON examples should make this task a bit easier for the end-user.

### DIFF
--- a/content/sensu-go/5.18/reference/assets.md
+++ b/content/sensu-go/5.18/reference/assets.md
@@ -395,6 +395,15 @@ spec:
 }
 {{< /highlight >}}
 
+{{< highlight ansible >}}
+- name: Minimal set of asset attributes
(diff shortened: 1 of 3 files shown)


cwjohnston:
Hi @tadeboro, thanks for taking the time to open this PR.

I think the idea has merit, but with our current priorities we aren't ready to expand the scope of these documents to include configuration management DSLs. 

For now I'll close this issue with our thanks and perhaps we will revisit the idea at a later date.

PR #2318, opened 2020-03-30
Update install Sensu links for 5.19

rgeniesse:
Update install-sensu with 10682 build number for 5.19


hillaryfraley:
@rgeniesse I'm not 100% sure the MIPS tar/zip links will work -- I guessed on the format based on links for other systems. I can help check them when the packages are published.

PR #2331, opened 2020-04-01
Add PowerShell example with -f flag to check reference

### LINKED ISSUE #2330: Document workaround for PowerShell scripts
### Description
PowerShell scripts do not return the right return code, as described in https://github.com/sensu/sensu/issues/1919.

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/sensu-go/latest/reference/checks/ ?

### Expected behavior
We should document the use of the `-f` flag, which appears to fix this issue.

### Current behavior
A user that doesn't specify this flag will only have return codes of `0` or `1`.

### Additional context

* Browser(s) and version(s): 
* OS:

hillaryfraley:
## Description
Adds a check example with PowerShell in command to check reference.

**Question**: Does the PowerShell workaround work with YAML? I included a yaml example, but the original issue only included JSON so I thought I should check.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2330

### DIFF
--- a/content/sensu-go/5.19/reference/checks.md
+++ b/content/sensu-go/5.19/reference/checks.md
@@ -1008,6 +1008,7 @@ spec:
 ### PowerShell script in check commands
 
 If you use a PowerShell script in your check command, make sure to include the `-f` flag in the command.
+The `-f` flag ensures that the proper exit code is passed into Sensu.
 For example:
 
 {{< language-toggle >}}
--- a/content/sensu-go/5.19/reference/checks.md
+++ b/content/sensu-go/5.19/reference/checks.md
@@ -1005,6 +1005,50 @@ spec:
 
 {{< /language-toggle >}}
 
+### PowerShell script in check commands
+
+If you use a PowerShell script in your check command, make sure to include the `-f` flag in the command.
+For example:
+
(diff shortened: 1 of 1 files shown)


content/sensu-go/5.19/reference/checks.md
```diff
@@ -1005,6 +1005,50 @@ spec:
 
 {{< /language-toggle >}}
 
+### PowerShell script in check commands
+
+If you use a PowerShell script in your check command, make sure to include the `-f` flag in the command.
```
rgeniesse, line 1010: May want to off some explanation into why the `-f` is required. My understanding of it is that it allows the proper exit code to be passed into Sensu.

hillaryfraley: Good point, thank you. I will add it.

hillaryfraley, line 1010:
```suggestion
If you use a PowerShell script in your check command, make sure to include the `-f` flag in the command.
The `-f` flag ensures that the proper exit code is passed into Sensu.
```

palourde (approved):
👍

PR #2339, opened 2020-04-03
Upgrade NodeJS and modules

cwjohnston:
##  Description

Upgrade Node version from 9 to 12, upgrade modules.

## Motivation and Context

Trying to resolve https://github.com/sensu/sensu-docs/network/alert/yarn.lock/minimist/open

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ matrix:
   include:
     - language: node_js
       nodejs:
-        - 9
+        - 12
 
       addons:
         apt:
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "grunt-sass": "^3.0.2",
     "node-sass": "^4.12.0",
     "toml": "^3.0.0",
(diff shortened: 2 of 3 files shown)


cwjohnston:
@hillaryfraley will you please do some quick manual checking of the test deployment to see if you detect any defects? I haven't noticed any yet.

cwjohnston:
From the diff it looks like this still leaves minimist 1.1.3 in play, but the advisory indicates we should be on 1.2.2 or later.

hillaryfraley (approved):
I clicked around quite a bit in https://sensu-docs-minify-upgra-wxzsxj.herokuapp.com/ and couldn't find anything wrong. I checked things like pages load properly, images show up, our clickable images open in another tab, internal and external links work, the embedded Katacoda scenarios load, and so on. I also checked the index page and dropdown docs selector to make sure the non-Go doc sets load properly.

As you mentioned, it looks like this doesn't solve the minimist problem, but it doesn't break the docs either.

PR #2392, opened 2020-04-23
Add sensuctl describe-type command

hillaryfraley:
## Description
Adds `sensuctl describe-type` section in sensuctl reference doc.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2376

PR #2418, opened 2020-05-01
Update resources lists in RBAC reference

hillaryfraley:
## Description
Adds resources to lists in RBAC reference: searches to namespaced list and provider, providers, and secrets to cluster-wide list

## Motivation and Context
https://github.com/sensu/sensu-docs/pull/2392#issuecomment-622416017

### DIFF
--- a/content/sensu-go/5.16/reference/rbac.md
+++ b/content/sensu-go/5.16/reference/rbac.md
@@ -211,14 +211,15 @@ You can access cluster-wide resources only by [cluster roles][21].
 
 | type | description |
 |---|---|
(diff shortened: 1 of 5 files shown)


palourde (approved):
👌

hillaryfraley:
@palourde if this is correct, I think providers and secrets should be propagated back to 5.17, and provider should be propagated back to 5.16, right? I think searches was added for 5.19, so it only needs to be propagated forward to 5.20.

palourde:
@hillaryfraley Yes, except for the the PostgreSQL event store, which was introduced in 5.10.0 (https://docs.sensu.io/sensu-go/latest/release-notes/#5-10-0-release-notes).

PR #2428, opened 2020-05-05
Add note about trace log level in Troubleshooting

### LINKED ISSUE #2407: Add note about trace log level in troubleshooting
### Description
We should add a note in the Troubleshooting log levels section to explain how Sensu handles the trace log level:
- Users can set the log level to trace, but we don't log at that level anywhere in the application, so it's functionally the same as debug level right now.
- Users can get to trace level by doing `--log-level trace` or by incrementing via SIGUSR1

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/sensu-go/5.20/guides/troubleshooting/#log-levels

### Expected behavior
The page should contain a note explaining how Sensu handles the log level

### Current behavior
As of 5.20, the increment info includes mention of the trace level but does not explain how users could end up at that log level.

### Corrections needed
Add a note explaining how Sensu handles the log level and how users could end up there.

hillaryfraley:
## Description
Adds note to explain that trace is the same as debug in Sensu.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2407

### DIFF
--- a/content/sensu-go/5.16/guides/troubleshooting.md
+++ b/content/sensu-go/5.16/guides/troubleshooting.md
@@ -46,6 +46,11 @@ sensu-agent start --log-level debug
 You must restart the service if you change log levels via configuration files or command line arguments.
 For help with restarting a service, see the [agent reference][5] or [backend reference][9].
 
+{{% notice note %}}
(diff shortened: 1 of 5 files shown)


hillaryfraley:
Hmm maybe I should say "The trace log level output is the same as debug in Sensu" instead.

PR #2430, opened 2020-05-07
Add example registration handler

hillaryfraley:
## Description
- Adds an example registration handler in the handlers reference.
- Updates link in registration events content in agent reference so that it points to the new registration handler.

## Motivation and Context
Jef noticed that our agent reference says the handlers reference includes a registration handler example, but it does not.

## Review Instructions
Once approved, is it appropriate to propagate this to the 5.16, 5.17, and 5.18 docs?


nixwiz (approved):
LGTM! 👍

hillaryfraley:
Thanks @nixwiz ! Would this handler work for previous supported Sensu versions? Specifically, 5.16, 5.17, and 5.18?

nixwiz:
> Thanks @nixwiz ! Would this handler work for previous supported Sensu versions? Specifically, 5.16, 5.17, and 5.18?

Yes, it should work.

PR #2436, opened 2020-05-08
Add web configuration documentation

### LINKED ISSUE #2432: Add section of detailing web configuration feature
### Description

Web Configuration will allow administrators to configure features and defaults in their dashboard.

An example 

```yaml
type: GlobalConfig
api_version: web/v1
metadata:
  name: default
spec:
  #
  # When one or more clusters are configured the default behaviour of the application is to omit
  # the details of the cluster the user is currently connected to. For purposes of debugging it can
  # be useful to enable.
  always_show_local_cluster: false

  #
  # Administrators can set global default preferences for all users. At this time they can tweak
  # the default theme users will see and the default amount of items a user can see on each page.
  default_preferences:
    page_size: 50
    theme: classic

  #
  # When a label or annotation contains a URL, the web app will attempt to turn it into a link or
  # and image. If administrators want they can configure which domains are valid/invalid targets.
  link_policy:
    # allow_list controls whether the list acts as an allow or deny list.
    # - as an allow list only matching URLs will be expanded
    # - as a deny list matching URLs are /not/ expanded but everything else is. 
    allow_list: true
    urls:
    - https://example.com
    - steamapp://34234234
    - //google.com
    - //*.google.com
    - //bob.local
```

hillaryfraley:
## Description
- Adds web configuration API doc
- Adds web configuration reference doc
- Adds web configuration guide in Dashboard section
- Adds `GlobalConfig` resource to sensuctl create types
- Fixes weight of other two Dashboard section docs so the new doc is in the correct order
- Updates commercial features doc to include web config info

## Motivation and Context
Closes #2432

### DIFF
--- a/content/sensu-go/5.20/api/webconfig.md
+++ b/content/sensu-go/5.20/api/webconfig.md
@@ -1,0 +1,328 @@
+---
+title: "Web configuration API"
+linkTitle: "Web Configuration API"
+description: "The Sensu web API provides HTTP access to the active web UI configuration. This reference includes examples for returning the active web configuration and adding or updating the web configuration. Read on for the full reference."
+version: "5.20"
+product: "Sensu Go"
+menu:
+  sensu-go-5.20:
+    parent: api
+---
+
+- [The `/globals` API endpoint](#the-globals-api-endpoint)
+  - [`/globals` (GET)](#globals-get)
+- [The `/globals/:globalconfig` API endpoint](#the-globalsglobalconfig-api-endpoint)
+  - [`/globals/:globalconfig` (GET)](#globalsglobalconfig-get)
+  - [`/globals/:globalconfig` (PUT)](#globalsglobalconfig-put)
+  - [`/globals/:globalconfig` (DELETE)](#globalsglobalconfig-delete)
+
+## The `/globals` API endpoints
+
+For more information about commercial features designed for enterprises, see [Get started with commercial features][1].
+
+### `/globals` (GET)
+
+The `/globals` API endpoint provides HTTP GET access to the global web UI configuration.
+
+#### EXAMPLE {#globals-get-example}
+
+The following example demonstrates a request to the `/web` API endpoint, resulting in a JSON array that contains the global web UI configuration.
+
+{{< highlight shell >}}
+curl -X GET \
+http://127.0.0.1:8080/api/enterprise/web/v1/globals \
+-H "Authorization: Bearer $SENSU_ACCESS_TOKEN" \
+-H 'Content-Type: application/json'
+
+HTTP/1.1 200 OK
+[
+  {
+    "type": "GlobalConfig",
+    "api_version": "web/v1",
+    "metadata": {
+      "name": "custom-web-ui",
+      "created_by": "admin"
+    },
+    "spec": {
+      "always_show_local_cluster": false,
+      "default_preferences": {
+        "page_size": 50,
+        "theme": "classic"
+      },
+      "link_policy": {
+        "allow_list": true,
+        "urls": [
+          "https://example.com",
+          "steamapp://34234234",
+          "//google.com",
+          "//*.google.com",
+          "//bob.local"
+        ]
+      }
+    }
+  },
+  {
+    "type": "GlobalConfig",
+    "api_version": "web/v1",
+    "metadata": {
+      "name": "new-web-ui",
+      "created_by": "admin"
+    },
+    "spec": {
+      "always_show_local_cluster": true,
+      "default_preferences": {
+        "page_size": 20,
+        "theme": "classic"
+      },
+      "link_policy": {
+        "allow_list": true,
+        "urls": [
+          "https://example.com"
+        ]
+      }
+    }
+  }
+]
+{{< /highlight >}}
+
+#### API Specification {#globals-get-specification}
+
+/web (GET)  | 
+---------------|------
+description    | Returns the list of global web UI configurations.
+example url    | http://hostname:8080/api/enterprise/web/v1/globals
+response type  | Map
+response codes | <ul><li>**Success**: 200 (OK)</li><li>**Error**: 500 (Internal Server Error)</li></ul>
+output         | {{< highlight shell >}}
+[
+  {
+    "type": "GlobalConfig",
+    "api_version": "web/v1",
+    "metadata": {
+      "name": "custom-web-ui",
+      "created_by": "admin"
+    },
+    "spec": {
+      "always_show_local_cluster": false,
+      "default_preferences": {
+        "page_size": 50,
+        "theme": "classic"
+      },
+      "link_policy": {
+        "allow_list": true,
+        "urls": [
+          "https://example.com",
+          "steamapp://34234234",
+          "//google.com",
+          "//*.google.com",
+          "//bob.local"
+        ]
+      }
+    }
+  },
+  {
+    "type": "GlobalConfig",
+    "api_version": "web/v1",
+    "metadata": {
+      "name": "new-web-ui",
+      "created_by": "admin"
+    },
+    "spec": {
+      "always_show_local_cluster": true,
+      "default_preferences": {
+        "page_size": 20,
+        "theme": "classic"
+      },
+      "link_policy": {
+        "allow_list": true,
+        "urls": [
+          "https://example.com"
+        ]
+      }
+    }
+  }
+]
+{{< /highlight >}}
+
+## The `/globals/:globalconfig` API endpoint {#the-globalsglobalconfig-api-endpoint}
+
+### `/globals/:globalconfig` (GET) {#globalsglobalconfig-get}
+
+The `/globals/:globalconfig` API endpoint provides HTTP GET access to global web UI configuration data, specified by configuration name.
+
+#### EXAMPLE {#globalsglobalconfig-get-example}
+
+In the following example, querying the `/globals/:globalconfig` API endpoint returns a JSON map that contains the requested [`:globalconfig` definition][1] (in this example, for the `:globalconfig` named `custom-web-ui`).
+
+{{< highlight shell >}}
+curl -X GET \
+http://127.0.0.1:8080/api/enterprise/web/v1/globals/custom-web-ui \
+-H "Authorization: Bearer $SENSU_ACCESS_TOKEN"
+
+HTTP/1.1 200 OK
+{
+  "type": "GlobalConfig",
+  "api_version": "web/v1",
+  "metadata": {
+    "name": "custom-web-ui",
+    "created_by": "admin"
+  },
+  "spec": {
+    "always_show_local_cluster": false,
+    "default_preferences": {
+      "page_size": 50,
+      "theme": "classic"
+    },
+    "link_policy": {
+      "allow_list": true,
+      "urls": [
+        "https://example.com",
+        "steamapp://34234234",
+        "//google.com",
+        "//*.google.com",
+        "//bob.local"
+      ]
+    }
+  }
+}
+{{< /highlight >}}
+
+#### API Specification {#globalsglobalconfig-get-specification}
+
+/globals/:globalconfig (GET) | 
+---------------------|------
+description          | Returns the specified global web UI configuration.
+example url          | http://hostname:8080/api/enterprise/web/v1/globals/custom-web-ui
+response type        | Map
+response codes       | <ul><li>**Success**: 200 (OK)</li><li> **Missing**: 404 (Not Found)</li><li>**Error**: 500 (Internal Server Error)</li></ul>
+output               | {{< highlight json >}}
+{
+  "type": "GlobalConfig",
+  "api_version": "web/v1",
+  "metadata": {
+    "name": "custom-web-ui",
+    "created_by": "admin"
+  },
+  "spec": {
+    "always_show_local_cluster": false,
+    "default_preferences": {
+      "page_size": 50,
+      "theme": "classic"
+    },
+    "link_policy": {
+      "allow_list": true,
+      "urls": [
+        "https://example.com",
+        "steamapp://34234234",
+        "//google.com",
+        "//*.google.com",
+        "//bob.local"
+      ]
+    }
+  }
+}
+{{< /highlight >}}
+
+### `/globals/:globalconfig` (PUT) {#globalsglobalconfig-put}
+
+The `/globals/:globalconfig` API endpoint provides HTTP PUT access to create and update global web UI configurations, specified by configuration name.
+
+#### EXAMPLE {#globalsglobalconfighooks-put-example}
+
+In the following example, an HTTP PUT request is submitted to the `/globals/:globalconfig` API endpoint to update the `custom-web-ui` configuration, resulting in an HTTP `200 OK` response and the updated configuration definition.
+
+{{< highlight shell >}}
+curl -X PUT \
+-H "Authorization: Bearer $SENSU_ACCESS_TOKEN" \
+-H 'Content-Type: application/json' \
+-d '{
+  "type": "GlobalConfig",
+  "api_version": "web/v1",
+  "metadata": {
+    "name": "custom-web-ui",
+    "created_by": "admin"
+  },
+  "spec": {
+    "always_show_local_cluster": false,
+    "default_preferences": {
+      "page_size": 50,
+      "theme": "classic"
+    },
+    "link_policy": {
+      "allow_list": true,
+      "urls": [
+        "https://example.com",
+        "steamapp://34234234",
+        "//google.com",
+        "//*.google.com",
+        "//bob.local"
+      ]
+    }
+  }
+}' \
+http://127.0.0.1:8080/api/enterprise/web/v1/globals/custom-web-ui
+
+HTTP/1.1 201 Created
+{{< /highlight >}}
+
+#### API Specification {#globalsglobalconfig-put-specification}
+
+/globals/:globalconfig (PUT) | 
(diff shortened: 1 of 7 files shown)


hillaryfraley:
Still possibly need a guide in the Dashboard content to describe this feature.

hillaryfraley:
@jamesdphillips I have some questions before I draft guide-type content:

1. How do I tell Sensu to use my custom web UI config?

2. If I have more than one custom web UI config, how do I specify which config to use? I can create and GET more than one config. How do I tell the web UI which one I want to use?

3. How do I specify which web UI configs should be used for which clusters?

4. Does an individual user's system settings override the web config? For example, if you set the mode to classic but a user's system is set to dark mode and they have their web UI set to pull from their system settings, will that user see classic or dark in Sensu's web UI?

5. If I send a completely blank PUT, Sensu adds a default for everything except theme. If you use a web UI config but it doesn't include a theme, what theme does the web UI use? Note that I mean with an empty spec, Sensu does not even add the `theme` key with a null default value...it just doesn't list theme at all. What happens in that case?

6. If i don't specify a list size, Sensu defaults to 0. How does a default list size of 0 work? Does that mean users won't see any items per page at all?

7. Can other users besides the admin create/update web UI configs?

8. What are the available values for theme? Is it only dark or classic?

9. I'm not sure what this note from https://github.com/sensu/sensu-enterprise-go/pull/965 means:
```
During the initial meeting when we discussed this feature, it was brought up that this API is an exception to the rule in that there should only ever be a single web config present. In lieu of a bigger refactor said limitation is preset in this PR, hopefully we can avoid any issues by documenting this limitation.
```
It sounds like this says you can only have one web UI config, but I was able to create many of them.
   9a. Does this note mean we want to tell people they should only create one web UI config (even though they technically can create more than one)?
   9b. What would be the effects of creating more than one?
   9c. If they should only create one web UI config, why do we say they can have a different config for different clusters?

10. If always_show_local_cluster is set to true, where are these cluster details displayed? And which details do they include (i.e., is it just the web UI config details for that cluster, or something else)?

jamesdphillips:
Sorry again for my lack of details. Great questions. 

> How do I tell Sensu to use my custom web UI config?

I think the suggested way to configure should be to use `sensuctl create`.

> If I have more than one custom web UI config, how do I specify which config to use? I can create and GET more than one config. How do I tell the web UI which one I want to use?

Similar to our Postgres adapter endpoint, if more than one configuration is present only the first will be honoured. So while, yes, you could have more than one it wouldn't make sense to.

> How do I specify which web UI configs should be used for which clusters?

Good question, not something I had put much thought into.

The configuration applies to the web app irrespective of the cluster they're currently viewing and is configuration used is provided by the cluster they connected to. So for instance, if they navigated to `https://cluster-a.sensu.my.org:3000/` the application would be configured using the values set on `cluster-a`.

If they're in a federated environment, we should probably suggest that they configure an etcd replicator for the resource. e.g..

```
api_version: federation/v1
type: EtcdReplicator
metadata:
  name: web_global_config
spec:
  ca_cert: /path/to/ssl/trusted-certificate-authorities.pem
  cert: /path/to/ssl/cert.pem
  key: /path/to/ssl/key.pem
  insecure: false
  url: http://127.0.0.1:2379
  api_version: web/v1
  resource: GlobalConfig
  replication_interval_seconds: 120
```

> Does an individual user's system settings override the web config? For example, if you set the mode to classic but a user's system is set to dark mode and they have their web UI set to pull from their system settings, will that user see classic or dark in Sensu's web UI?

Yes!

> If I send a completely blank PUT, Sensu adds a default for everything except theme. If you use a web UI config but it doesn't include a theme, what theme does the web UI use? Note that I mean with an empty spec, Sensu does not even add the theme key with a null default value...it just doesn't list theme at all. What happens in that case?

The global configuration overrides the apps own hardcoded defaults, the administrator doesn't provide a value the hardcoded default is used. For the theme specifically, the "sensu" theme is used.

> If i don't specify a list size, Sensu defaults to 0. How does a default list size of 0 work? Does that mean users won't see any items per page at all?

Similar to the answer above. It will fallback to the hardcoded default of 25.

> Can other users besides the admin create/update web UI configs?

Cluster admins and anyone with access to the `webconfig` resource.

> What are the available values for theme? Is it only dark or classic?

It can be one of.. `sensu`, `classic`, `uchiwa`, `tritanopia`, & `deuteranopia`.

Configuring dark mode is not supported in the initial release of the web config feature.

> I'm not sure what this note from sensu/sensu-enterprise-go#965 means. It sounds like this says you can only have one web UI config, but I was able to create many of them.

Noted above but while you can create many, only the first one will be used.

> If always_show_local_cluster is set to true, where are these cluster details displayed? And which details do they include (i.e., is it just the web UI config details for that cluster, or something else)?

If you have the creatively named `cluster-a`, `cluster-b`, and `cluster-c` and you connect to the web interface on `cluster-b`, in the context switcher you'll see a heading named `local-cluster` under which will appear the namespaces on `cluster-b`.

It's more for helping debug (eg. `cluster-b` isn't show up in the switcher) or when you have a cluster that you don't want to advertise as part of the federation (eg. you can directly connect to `cluster-d` but you don't want the other clusters to be aware of it's presence.)

hillaryfraley:
@jamesdphillips I think this is ready for review, with a couple more questions.

1. Under "Local cluster settings for debugging" in https://github.com/sensu/sensu-docs/pull/2436/files#diff-7de5e7cf1ce9e4e9760357d323b2ec53, I'm not sure how to explain what it tells someone if a particular cluster missing from the namespace switcher. I am definitely missing something here: how can you open the web UI for a specific cluster, and still have that cluster be missing from the namespace switcher list?

2. I don't understand what `always_show_local_cluster` does exactly. In my web UI, nothing seems to change whether I set it to true or false. In the namespace switcher, I see `local-cluster` followed by my three namespaces either way.

3. When you say Sensu will use the first configuration, I assumed you mean the earliest-created -- is that correct?
3a. Does that also mean earliest created OR updated? For example, you create config1 and config2. You created config1 first, so it's the earliest. But then you update config1. In this scenario, does Sensu use config1 or config2?

4. I wonder if it will be confusing that the default page size is 25, but if you forget to specify a page size altogether, the response lists a value of `0`. Shouldn't it list `25` instead?

5. Are there any other examples we should include in the reference?

content/sensu-go/5.20/api/webconfig.md
```diff
@@ -0,0 +1,286 @@
...
+linkTitle: "Web UI Configuration API"
+description: "The Sensu web API provides HTTP access to the active web UI configuration. This reference includes examples for returning the active web configuration and adding or updating the web configuration. Read on for the full reference."
+version: "5.20"
+product: "Sensu Go"
+menu:
+  sensu-go-5.20:
+    parent: api
+---
+
+- [The `/globals` API endpoint](#the-globals-api-endpoint)
```
jamesdphillips: I apologize, I've last minute changed the endpoint. It's now `http://127.0.0.1:8080/api/enterprise/web/v1/globals`.

See: https://github.com/sensu/sensu-enterprise-go/pull/1026

hillaryfraley: @jamesdphillips sorry, so did v1 change to v2 and globals changed to config?

hillaryfraley: ^ I changed it to `...web/v2/config` so just let me know if that's incorrect. Thank you for the heads up!

jamesdphillips:
@hillaryfraley 

> Under "Local cluster settings for debugging" in https://github.com/sensu/sensu-docs/pull/2436/files#diff-7de5e7cf1ce9e4e9760357d323b2ec53, I'm not sure how to explain what it tells someone if a particular cluster missing from the namespace switcher. I am definitely missing something here: how can you open the web UI for a specific cluster, and still have that cluster be missing from the namespace switcher list?

> I don't understand what always_show_local_cluster does exactly. In my web UI, nothing seems to change whether I set it to true or false. In the namespace switcher, I see local-cluster followed by my three namespaces either way.

Sorry, I do realize how confusing this feature is.

The `always_show_local_cluster` only applies to federated environments. Based on configuration the context switcher could be organized three ways.

1. In a non-federated environment, you will only see a `local-cluster` heading with any namespaces you have configured.
1. in a federated environment, you will see a heading for _each_ cluster you have configured. eg. if you have three clusters configured `cluster-a`, `cluster-b`, and `cluster-c` and you open the switcher you will see a heading for each and namespaces under each heading.
1. in a federated environment with the `always_show_local_cluster` flag set to true, you will see a heading for each cluster PLUS the `local-cluster` heading.

I see it's potential utility as:

* In a federated environment where, due to error / misconfiguration / network issues one cluster isn't showing up in the context switcher. The administration could enable `always_show_local_cluster` and directly connect to the cluster that isn't displaying.
* Say we have a federated environment with three clusters configured, let's call them A, B, and C. Then imagine we have a fourth cluster D but we don't want to advertise it. To do this we could set up an etcd replicator so that cluster D has the configuration we want and to avoid being advertised we could simply avoid adding an entry to the [cluster API](https://docs.sensu.io/sensu-go/latest/api/cluster/). This way cluster D will not show up in the context switcher.
Of course this is a double edge sword because even if we now directly connect to cluster D, the context switcher will only allow us to switch between clusters A through C. Enabling `always_show_local_cluster` would allow us to manage the "local-cluster" D.

I'll also add: if it's too tricky to communicate how this feature works, I would be completely OK with shelving documenting it for now. I do feel as if its utility is very limited and only really applies (at this point) to theoretical edge cases.

> When you say Sensu will use the first configuration, I assumed you mean the earliest-created -- is that correct? Does that also mean earliest created OR updated? For example, you create config1 and config2. You created config1 first, so it's the earliest. But then you update config1. In this scenario, does Sensu use config1 or config2?

Great question, and to be honest I am not sure. I think in the scenario you've painted the answer is `config2` but I'll have to double check.

For what it's worth, it might not even matter. The most important point is that administrators avoid doing it in the first place.

> I wonder if it will be confusing that the default page size is 25, but if you forget to specify a page size altogether, the response lists a value of 0. Shouldn't it list 25 instead?

This is a good point and maybe this is an unfortunate mistake on my part. Similar to the theme being `""` if you don't provide a value, a page size of `0` is the [zero value](https://tour.golang.org/basics/12/) for the field. The defaults live within the web application itself and apply if they receive the zero value.

> Are there any other examples we should include in the reference?

What you've included looks great :)

hillaryfraley:
Thank you for all of your clarifications, @jamesdphillips ! I think this is actually ready for your review now.

jamesdphillips (approved):
🎉

content/sensu-go/5.20/api/webconfig.md
```diff
@@ -0,0 +1,286 @@
+---
+title: "Web UI configuration API"
+linkTitle: "Web UI Configuration API"
+description: "The Sensu web API provides HTTP access to the active web UI configuration. This reference includes examples for returning the active web configuration and adding or updating the web configuration. Read on for the full reference."
```
hillaryfraley:
```suggestion
description: "The Sensu web configuration API provides HTTP access to the global web UI configuration. This reference includes examples for returning the global web UI configuration and adding or updating the web UI configuration. Read on for the full reference."
```

content/sensu-go/5.20/api/webconfig.md
```diff
@@ -0,0 +1,286 @@
...
+---
+
+- [The `/config` API endpoint](#the-config-api-endpoint)
+  - [`/config` (GET)](#config-get)
+- [The `/config/:globalconfig` API endpoint](#the-configglobalconfig-api-endpoint)
+  - [`/config/:globalconfig` (GET)](#configglobalconfig-get)
+  - [`/config/:globalconfig` (PUT)](#configglobalconfig-put)
+  - [`/config/:globalconfig` (DELETE)](#configglobalconfig-delete)
+
+## The `/config` API endpoint
```
hillaryfraley:
```suggestion
**COMMERCIAL FEATURE**: Access web UI configuration in the packaged Sensu Go distribution.
For more information, see [Get started with commercial features][1].

## The `/config` API endpoint
```

content/sensu-go/5.20/api/webconfig.md
jamesdphillips: Whoops-- I missed this comment. It is `/api/enterprise/web/v1/config`.

Sorry for the confusion.

content/sensu-go/5.20/api/webconfig.md
```diff
@@ -0,0 +1,286 @@
...
+
+For more information about commercial features designed for enterprises, see [Get started with commercial features][1].
```
hillaryfraley:
```suggestion
```

content/sensu-go/5.20/api/webconfig.md
```diff
@@ -0,0 +1,286 @@
...
+
+## The `/config/:globalconfig` API endpoint {#the-configglobalconfig-api-endpoint}
+
+### `/config/:globalconfig` (GET) {#configglobalconfig-get}
+
+The `/config/:globalconfig` API endpoint provides HTTP GET access to global web UI configuration data, specified by configuration name.
+
+#### EXAMPLE {#configglobalconfig-get-example}
+
+In the following example, querying the `/config/:globalconfig` API endpoint returns a JSON map that contains the requested [`:globalconfig` definition][1] (in this example, for the `:globalconfig` named `custom-web-ui`).
```
hillaryfraley:
```suggestion
In the following example, querying the `/config/:globalconfig` API endpoint returns a JSON map that contains the requested `:globalconfig` definition (in this example, for the `:globalconfig` named `custom-web-ui`).
```

content/sensu-go/5.20/dashboard/webconfig.md
```diff
@@ -0,0 +1,104 @@
...
+
+Web UI configuration allows you to define certain display options for the Sensu [web UI][3], such as which web UI theme to use, the number of items to list on each page, and which URLs and linked images to expand.
+You can define a single custom web UI configuration to federate to all, some, or only one of your clusters.
+
+## Create a web UI configuration
+
+Use the [web UI configuration API][2] or [sensuctl create][5] to create a `GlobalConfig` resource.
+The [web UI configuration reference][4] describes each attribute you can configure in the `GlobalConfig` resource.
+
+{{% notice important %}}
```
hillaryfraley:
```suggestion
{{% notice note %}}
```

content/sensu-go/5.20/reference/webconfig.md
```diff
@@ -0,0 +1,248 @@
...
+
+**COMMERCIAL FEATURE**: Access web UI configuration in the packaged Sensu Go distribution.
+For more information, see [Get started with commercial features][1].
+
+Web UI configuration allows you to define certain display options for the Sensu [web UI][3], such as which web UI theme to use, the number of items to list on each page, and which URLs and linked images to expand.
+You can define a single custom web UI configuration to federate to all, some, or only one of your clusters.
+
+{{% notice note %}}
+**NOTE**: Each cluster should have only one web configuration.
+If you provide more than one, Sensu will automatically use the web UI configuration that was created or updated first.
```
hillaryfraley, line 25:
```suggestion
**NOTE**: Each cluster should have only one web configuration.
```

hillaryfraley, line 25:
```suggestion
**NOTE**: Each cluster should have only one web configuration.
```

hillaryfraley:
```suggestion
```

content/sensu-go/5.20/reference/webconfig.md
```diff
@@ -0,0 +1,247 @@
...
+required     | Required for web UI configuration in `wrapped-json` or `yaml` format.
+type         | String
+example      | {{< highlight shell >}}"type": "GlobalConfig"{{< /highlight >}}
+
+api_version  | 
+-------------|------
+description  | Top-level attribute that specifies the Sensu API group and version. For web UI configuration in this version of Sensu, the api_version should always be `web/v2`.
+required     | Required for web UI configuration in `wrapped-json` or `yaml` format.
+type         | String
+example      | {{< highlight shell >}}"api_version": "web/v2"{{< /highlight >}}
```
hillaryfraley:
```suggestion
description  | Top-level attribute that specifies the Sensu API group and version. For web UI configuration in this version of Sensu, the api_version should always be `web/v1`.
required     | Required for web UI configuration in `wrapped-json` or `yaml` format.
type         | String
example      | {{< highlight shell >}}"api_version": "web/v1"{{< /highlight >}}
```

content/sensu-go/5.20/reference/webconfig.md
```diff
@@ -0,0 +1,247 @@
...
+- Details for the local cluster will not be displayed
+- Each page will list 50 items
+- The web UI will use the classic theme
+- Expanded links and images will be allowed for the listed URLs
+
+{{< language-toggle >}}
+
+{{< highlight yml >}}
+type: GlobalConfig
+api_version: web/v2
```
hillaryfraley:
```suggestion
api_version: web/v1
```

content/sensu-go/5.20/reference/webconfig.md
```diff
@@ -0,0 +1,247 @@
...
+    - steamapp://34234234
+    - //google.com
+    - //*.google.com
+    - //bob.local
+{{< /highlight >}}
+
+{{< highlight json >}}
+{
+  "type": "GlobalConfig",
+  "api_version": "web/v2",
```
hillaryfraley:
```suggestion
  "api_version": "web/v1",
```

Issue #2443, opened 2020-05-11
Navigation enhancements needed


hillaryfraley:
### Description
We would like to understand more about the capabilities of our current theme in context of some changes we want to make to the way our documentation is organized, nested, and displayed. 

With that in mind, we have these questions:

1. Can we add more nested levels in our left-nav menu? For example, could we nest more pages of content under a page like https://docs.sensu.io/sensu-go/latest/sensuctl/reference/?

2. Can we configure the docs site so that clicking a menu item opens a page with content? For example, could we turn the "Sensuctl CLI" menu item (in the left nav of our docs https://docs.sensu.io/sensu-go/latest/) into a page that opens in the main panel when clicked and displays content (but still includes nested content as well)?

3. Can we add a menu on the right side of each docs page that picks up and links the H2 and H3 headers on each page? Essentially, this would be an automatically generated table of contents for each individual page. See https://grafana.com/docs/grafana/latest/installation/requirements/ (right side, under "On this page:") for an example.

4. Can we add a global widget somewhere near the top on every page that would include our invitations to contribute (e.g. "Edit this page" and "Open a docs issue")? Again https://grafana.com/docs/grafana/latest/installation/requirements/ has a good example just above the "On this page:" menu.

Can we incorporate any of these changes and still keep our current docs theme?

### Category

This is an issue with:

- [ ] bug (bug in site build, layout, or functionality)
- [x] enhancement (add new or updated site functionality)

### Affected pages
All

### Additional context
If we can achieve these changes with our current docs theme, we can create individual issues to make PRs against. For now, we want to understand whether any of these changes are possible with our current docs theme.

rnevius:
So Sensu isn't actually using much of the hugo-material-docs theme anymore. Specifically, the theme only contains 10 lines of HTML and 2 images that aren't covered by the custom layouts and styles in this app. There's so much custom functionality baked into sensu-docs now that nearly all original theme layouts, shortcodes, and partials have been overridden. In effect, Sensu has its own Hugo theme at this point. 

To demonstrate, I have created [a new branch](https://github.com/sensu/sensu-docs/compare/demo/no-theme) that shows how easy it would be to completely drop hugo-material-docs as a dependency (I recommend doing this anyway at this point). 

So, with this in mind, we should determine whether or not the features requested in this issue can be added to our custom theme components.

[...]

PR #2451, opened 2020-05-13
Remove dependency on hugo-material-docs theme

cwjohnston:
## Description

Removes dependency on a third party Hugo theme.

## Motivation and Context

As described by Ryan in #2443, we aren't using very much of the code from hugo-material-docs theme, so this change imports what little we are still using so we can be rid of the external dependency all together.

## Review Instructions

Review preview deployment for any irregularities compared to production.


hillaryfraley (approved):
Looks good! I clicked through the pages and tried out platform switchers and the like -- I didn't find any problems.

hillaryfraley:
I think the newer PR https://github.com/sensu/sensu-docs/pull/2466 from @rnevius also removes the Hugo Material docs theme

cwjohnston:
Closing in favor of #2466

PR #2461, opened 2020-05-15
Update cloud_provider, memory_percent, and cpu_percent attribute descriptions

hillaryfraley:
## Description
- In 5.19 and 5.20 docs, specifies that cloud_provider values are returned only for EC2, GCP, and Azure.
- In 5.20 doc, adds notes for memory_percent and cpu_percent to explain that these attributes are only supported on Linux and macOS (not Windows).
- In 5.20 doc, explains how float works for memory_percent and cpu_percent and corrects examples.

## Motivation and Context
Corrections discovered in the course of 5.20.1 work.


palourde (approved):
👌

PR #2468, opened 2020-05-18
Update supported platforms to reflect accurate packages and binaries

hillaryfraley:
## Description
Updates supported platforms. doc to correctly reflect which Sensu provides for each platform.


amdprophet (approved):
Looks great to me!

PR #2484, opened 2020-05-28
Add use cases for entity system attributes in asset reference

### LINKED ISSUE #2378: Explain importance of entity.system attributes in asset reference
### Description
entity.system attributes like arch, os, platform, libc_type, etc. are important in terms of how a particular asset can be used. For example, some assets will be compiled for a specific libc.

The asset reference should include more information about the impact of entity.system attributes , including example use cases for entity.system attributes.

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/sensu-go/latest/reference/assets

### Expected behavior
The asset reference describes how different entity.system attributes affect asset usage.

### Current behavior
There is some explanation in the `filters` attribute description under https://docs.sensu.io/sensu-go/latest/reference/assets/#spec-attributes, but it does not include examples or details.

### Corrections needed
Add a section to the assets reference to describe example use cases for entity.system attributes like arch, os, platform, platform_family, arm_version, and libc_type.

### Additional context
This is a good example of an explanation of the impact of system attributes:
https://code.visualstudio.com/docs/remote/linux

hillaryfraley:
## Description
Adds section to explain use cases for entity.system attributes in https://docs.sensu.io/sensu-go/latest/reference/assets/

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2378

## Review instructions
I think this needs some detail about:
1. How the agent evaluates the entity.system attributes against the asset filters.
2. What happens if you try to install an asset compiled for a specific system and your configuration does not match. For example, does it fail to install the asset? Install the asset but it doesn't work? Return an error?


nixwiz:
The details about how it's evaluated are just above it under the filters spec `If multiple expressions are included, each expression must return true for Sensu to install the asset.` This also answers question 2, to an extent.  An asset can have several builds associated with it, with varying filters to match different systems (normally done around OS, platform, and architecture).  But if none match then the asset doesn't get downloaded and used and whatever is referencing the asset will fail to run.

hillaryfraley:
@nixwiz thank you. I added some updates based on your explanation.

I think I'm still not quite understanding it correctly -- I am thinking of this as something the asset creator would add in the asset definition so that they (or others who use the asset) would not try to use an asset that won't work on their systems. But that doesn't seem right (sounds like there is no warning or error, just an absence of the check/handler/filter they were expecting).

content/sensu-go/5.20/reference/assets.md
```diff
@@ -368,6 +371,25 @@ example      | {{< highlight shell >}}
...
+
+Use the [entity.system attributes][10] in asset [filters][42] to specify which systems and configurations the asset can be used with.
+
+For example, if your asset is compiled for Linux AMD64 and 386 architectures, you can add filters for `os` and `arch` to your asset:
+
+{{< highlight shell >}}
+"filters": ["entity.system.os=='linux'", "entity.system.arch=='amd64'", "entity.system.arch=='386'"] 
+{{< /highlight >}}
+
+In this example, for Linux systems with either AMD64 or 386 architectures, the `os` filter expression and at least one of the `arch` filter expressions will evaluate as `true`.
```
nixwiz: All of these are AND'ed together and all have to be true.  Plus you will only ever have one arch, therefore this example would never work.

nixwiz:
The asset creator would provide the asset definition either via examples in the README for their project or in the Bonsai configuration for their asset if it's shared there.  If you have to develop your own filters for asset definitions, most of the time you should be able to glean most of the information from the asset tar file name like these:

```
sensu-influxdb-handler_3.1.2_darwin_386.tar.gz
sensu-influxdb-handler_3.1.2_darwin_amd64.tar.gz
sensu-influxdb-handler_3.1.2_windows_amd64.tar.gz
sensu-influxdb-handler_3.1.2_linux_armv7.tar.gz
sensu-influxdb-handler_3.1.2_linux_arm64.tar.gz
sensu-influxdb-handler_3.1.2_linux_386.tar.gz
sensu-influxdb-handler_3.1.2_linux_amd64.tar.gz
```

hillaryfraley:
@nixwiz thank you. I talked about this a bit with Cameron, and made some updates based on his explanations.

nixwiz:
And it looks fantastic!

hillaryfraley:
@nixwiz Thank you for helping me out!

PR #2406, opened 2020-04-28
Add log level increment commands

hillaryfraley:
## Description
Adds commands to increment log level in troubleshooting section (https://docs.sensu.io/sensu-go/latest/guides/troubleshooting/#log-levels)

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2394

## Review Instructions
@echlebek we have not documented the `trace` log level -- should it be added to the table here: https://docs.sensu.io/sensu-go/latest/guides/troubleshooting/#log-levels

PR #2490, opened 2020-06-02
Change agent default log level to info

hillaryfraley:
## Description
- Removes restart instruction from https://docs.sensu.io/sensu-go/5.21/guides/troubleshooting/#increment-log-level-verbosity
- In https://docs.sensu.io/sensu-go/5.21/reference/agent/#configuration-summary table, changes default log level to info rather than warn

## Motivation and Context
Support 5.21 release. See https://github.com/sensu/sensu-go/pull/3817 for details.


hillaryfraley:
@ccressent looks like we just added the note that you wanted to remove from the Troubleshooting guide in a 5.20 PR: https://github.com/sensu/sensu-docs/pull/2406/files

Can you confirm that it is no longer necessary to restart the service after incrementing log level verbosity as of 5.21?

ccressent:
> Can you confirm that it is no longer necessary to restart the service after incrementing log level verbosity as of 5.21?

Let me try to clarify. There are 2 main ways one can change the log level:
1. Change the relevant bit in the configuration file or change the `--log-level` option passed on the command line
2. Send the running process predefined signals (`USR1` and `USR2`) that it interprets as a runtime change of log level

For 1., in both cases (config file or command line arg; they are logically equivalent), the process needs to be restarted because it only assesses the log level it should be using during startup.

For 2. An already running process receives the signal, alters its behaviour, and keeps running.

Phrased another way, it is always necesarry to restart the process if you alter the log level with 1. and it is never necessary to restart the process if you alter the log level with 2.

On a similar note, if you change the log level using 2. and then restart the process, you effectively discard your changes since the process will restart with whatever log level is given in the configuration file or command line arguments!

hillaryfraley:
Got it, thank you for explaining @ccressent -- I need to remove that in 5.20 too.

PR #2521, opened 2020-06-11
add "client auth" to the certificate usages for backend

cwjohnston:
## Description

Add "client auth" to the allowed certificate usages for backend profile in CA configuration of Generate Certificates guide.

## Motivation and Context

Make backend certificates valid for use when authenticating etcd peer via certificates.


hillaryfraley (approved):
LGTM! 👍

PR #2544, opened 2020-06-16
Add Hugo 0.72.0 and floating Table of Contents

### LINKED ISSUE #2536: Floating TOC broke platform switcher
### Description
I reverted https://github.com/sensu/sensu-docs/pull/2466 and https://github.com/sensu/sensu-docs/pull/2532 because I found another issue after they were merged: the platform switcher didn't seem to work.
platform-switcher

I restored the feature/navigation-enhancements branch in hopes that the issue can be fixed there.

### Category

This is an issue with:

- [x] bug (bug in site build, layout, or functionality)
- [ ] enhancement (add new or updated site functionality)

### Affected pages
Pages that include the platform switcher, like https://docs.sensu.io/sensu-go/latest/installation/install-sensu/#install-the-sensu-backend

### Expected behavior
When a reader clicks one of the platform tabs, the page should display the corresponding code for that platform.

### Current behavior
When a reader clicks any of the platform tabs except the first, the place where the corresponding code should be displayed is blank.

Note that I reverted the floating TOC change in https://github.com/sensu/sensu-docs/pull/2533, so the published docs site currently works as expected. The behavior described here is shown in the gif above.

### Changes needed
(if possible) update branch feature/navigation-enhancements so the change to add the floating TOC does not break the platform switcher

rnevius:
## Description

A huge PR that makes me nauseous, but is simply the following:

– Upgrades Hugo to 0.72.0 to support new Table of Contents feature (and Goldmark renderer)
– Adds a floating Table of Contents to content
– Enables `toc: false` in front matter for disabling Table of Contents on a case-by-case basis
– Adds a `{{< code >}}` shortcode to fix `{{< highlight >}}` problems due to the Hugo upgrade

## Motivation and Context

Everything from #2533 is included. The important difference is that the `{{< highlight >}}` shortcode has been wrapped by a new `{{< code >}}` shortcode to fix #2536. Prior to this PR, Sensu was using the `{{< highlight >}}` shortcode in an unsupported way, by passing unknown languages to the shortcode (e.g. `{{< highlight "Ubuntu/Debian" >}}`). This worked in prior versions of Hugo, but the Hugo team has fixed the shortcode so that it only accepts supported languages (supported languages can be seen [here](https://gohugo.io/content-management/syntax-highlighting/#list-of-chroma-highlighting-languages)). As a result, we needed a way to pass custom code block names to the platform switcher components, while still rendering code blocks with supported languages.

I have programmatically replaced all instances of `{{< highlight >}}` with `{{< code >}}`. The new spec for this shortcode is as follows: `{{< code supported_lang_name "Optional Display Name" >}}` (for example, `{{< code bash "Ubuntu/Debian" >}}`). Please let me know if you have any questions about the shortcode: https://github.com/sensu/sensu-docs/blob/deb91f780cee46a4f67b3045ad5763dd1424bdaf/layouts/shortcodes/code.html

A huge benefit of the new shortcode is that we have MUCH better syntax highlighting in most cases. See https://docs.sensu.io/sensu-go/latest/installation/install-sensu/#2-configure-and-start compared to the same page on staging, for example.

Resolves #2536 
Fixes bugs introduced by #2533


hillaryfraley (approved):
I have tested everything I can think to test, and I can't find any problems this time. Thank you -- here we go...!

hillaryfraley:
Tentative 🎉

hillaryfraley:
And 2 hours later, a much more confident 🎉  
Thank you, @rnevius !

PR #2568, opened 2020-06-23
Fixing monitoring guide

### LINKED ISSUE #846: Guide: Monitoring Sensu with Sensu

asachs01:
## Description

Provides a draft of monitor the monitor but with some corrections.

## Motivation and Context

Closes #846


hillaryfraley:
A community member mentioned that the example check in this PR won't work because it’ll only error on a non-200 response. This is based on a clarification from Engineering about the Health API endpoint: We do not interpret the health response, meaning we do not change the response code based on the response we get from etcd. See https://github.com/sensu/sensu-go/blob/master/backend/apid/routers/health.go#L38-L54. That endpoint only returns either 200 (OK) or 400 [which indicates the timeout query parameter provided in the request is invalid (e.g. not a number)].

asachs01:
@cwjohnston This should be ready for review

agoddard (approved):
lgtm! 👍 
* @asachs01 and I discussed updating the http check after publishing

PR #2569, opened 2020-06-23
Add silenced key to spec description table in events reference doc

### LINKED ISSUE #2567: Event reference does not document spec.silenced
### Description

Event reference `spec` does not describe the `silenced` attribute.

### Category

This is an issue with:

### Affected pages

Event reference

### Corrections needed

Add `silenced` attribute under `spec` attributes, describe the conditions under which it is present.

### Additional context

When there are no silencing entries matching a particular event, the `silenced` key will have a null value. In general API responses will omit keys with null values. As such, the `silenced` key is only expected to be present in events when one or more silencing entries match the event.

hillaryfraley:
## Description
Adds `silenced` to spec description table in events reference.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2567

## Review Instructions
I'm not sure if I picked a good example with "webserver" -- if there's a better example to use, I can replace it before propagating to all versions.


content/sensu-go/5.21/reference/events.md
```diff
@@ -570,7 +576,16 @@ occurrences_watermark | |
...
 required     | false
 type         | Integer greater than 0
-example      | {{< code shell >}}"occurrences_watermark": 1{{< /code >}}\
+example      | {{< code shell >}}"occurrences_watermark": 1{{< /code >}}
+
+silenced     | |
+-------------|------
+description  | Array of silencing entries that match the event. If no silencing entries match the event, the `silenced` key will have a null value. API responses omit keys with null values, so `silenced` will only appear for events when one or more silencing entries match the event.
```
cwjohnston:
```suggestion
description  | Array of silencing entries that match the event. The `silenced` key is only present for events where one or more silencing entries matched the event at time of processing. If the `silenced` key is not present in an event, the event was not silenced at the time of processing.
```

cwjohnston (changes requested):

PR #2572, opened 2020-06-25
Move migrate from Sensu Core to top level of docs site

hillaryfraley:
## Description
- Moves content about migrating from Sensu Core to Sensu Go to the top level of the docs site
- Updates ports content in migration guide
- Updates all links in docs for migrate doc

## Motivation and Context
Docs IA project


hillaryfraley:
Redirects are not working -- need to fix or remove

Edit: I'm removing the redirects. I don't think there are any external links to the migrate section in the Update doc.

cwjohnston:
@hillaryfraley I think this looks good but I’m questioning whether this new doc belongs under `Installation` instead of being a top-level document. WDYT?

hillaryfraley:
@cwjohnston I agree it should eventually be moved out of the top level (we discussed moving it into another category a bit later, when customers are finished migrating and the Core archive is removed). Do you think it should be in Installation now?

PR #2580, opened 2020-06-29
Fix the description of spec.client.version attribute in VaultProvider reference

cwjohnston:
## Description

Makes the description of .spec.client.version attribute more accurate.

## Motivation and Context

In https://sensu.slack.com/archives/C60EEQFH8/p1593202372412800 we discussed the issue a customer was facing in getting VaultProvider working. Through this discussion we learned that the `spec.client.version` attribute is actually specifying the version of the key/value store engine, not the version of the Vault HTTP API.


cwjohnston:
messed up this PR, will resubmit.

PR #2581, opened 2020-06-29
Fix/vault provider client version description

cwjohnston:
## Description

Corrects the description of `spec.client.version` attribute in VaultProvider reference

## Motivation and Context

In https://sensu.slack.com/archives/C60EEQFH8/p1593202372412800 we discussed the issue a customer was facing in getting VaultProvider working. Through this discussion we learned that the `spec.client.version` attribute is actually specifying the version of the key/value store engine, not the version of the Vault HTTP API.


hillaryfraley (approved):
Looks good to me! Thank you for catching and fixing this.

PR #2604, opened 2020-07-10
Fix table markdown for occurrences in events reference

cwjohnston:
## Description

Fixes markdown for [occurrences table in the event reference](https://docs.sensu.io/sensu-go/latest/reference/events/#occurrences-and-occurrences-watermark)

## Motivation and Context

I'm not sure if rendering this table broke when we upgraded Hugo, but here's what I'm seeing currently:

2020-07-10 at 3 11 PM

## Review Instructions

Table should look more like this in the preview deployment:

2020-07-10 at 3 13 PM


hillaryfraley (approved):
👍 thank you

PR #2608, opened 2020-07-13
Enable clicking on top-level menu items

rnevius:
## Description

Adds `href` attributes to top-level menu items so that these items can contain content. Clicking the chevron icons still expands the menus as usual.

## Motivation and Context

From #2443:

> Can we configure the docs site so that clicking a menu item opens a page with content? For example, could we turn the "Sensuctl CLI" menu item (in the left nav of our docs https://docs.sensu.io/sensu-go/latest/) into a page that opens in the main panel when clicked and displays content (but still includes nested content as well)?

## Review Instructions

Add any content you want to top-level `_index.md` templates.

### DIFF
--- a/layouts/partials/nav.html
+++ b/layouts/partials/nav.html
@@ -34,8 +34,6 @@
           {{ range $y, $x := sort $v "Weight" }}
 
           <!-- Determine if one the children of this menu is active so we can apply the proper styling -->
-            {{ $isCurrent := ($currentNode.HasMenuCurrent $k $x) }}
-
             <!-- if the menu has a child menu -->
             {{ if .HasChildren }}
               {{ if (in $k $version) }}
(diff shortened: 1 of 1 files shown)


hillaryfraley:
@rnevius thank you! This works exactly as I'd hoped.

One question: if I want to include the floating TOC in the right margin on the `_index.md` pages, it looks like all I need to do is change the layout to "single" -- is that the right way to make sure the floating TOC displays on the page in this case?

rnevius:
@hillaryfraley correct! The "single" page template is the one that includes the Table of Contents partial.

hillaryfraley (approved):
Approved! This is great. I'm going to hold on merging it until I have drafted the content I want to add to the index pages (hopefully by the end of the day).

Many thanks, @rnevius !

PR #2680, opened 2020-08-14
replace envvar reference  with valid token substitution in example

jspaleta:
## Description
The use of envvar in the asset header examples is malformed and does not work in practice.
I've replaced them with a token substitution example that I have tested locally using 6.0

## Motivation and Context
Working examples are great

## Review Instructions
We need to figure out how far back in the docs versions this needs to be applied. 
Asset token substitution is somewhat new.. in that it was not a feature in 5.0. 
Need to find when it was added in the release notes.

### DIFF
--- a/content/sensu-go/5.19/reference/assets.md
+++ b/content/sensu-go/5.19/reference/assets.md
@@ -251,7 +251,6 @@ example (multiple builds)     | {{< code shell >}}"spec": {
       }
   ],
   "headers": {
-    "Authorization": "Bearer $TOKEN",
     "X-Forwarded-For": "client1, proxy1, proxy2"
   }
 }{{< /code >}}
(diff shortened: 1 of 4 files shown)


jspaleta:
Okay it looks like 5.21 and 5.20 can have these changed applied.
5.19 however.. i think the header examples are just broken.  We don't have token substitution in 5.19 according to the release notes.. and i'm pretty sure the envvar $TOKEN example has never worked.    It definitely does not work in 6.0

jspaleta:
Okay I've updated 5.19  to remove the problematic header examples
5.20 5.21 and 6.0  all have the header examples updated to use token substitution with default fallback

hillaryfraley (approved):

PR #2688, opened 2020-08-20
Clarify health API HTTP response codes and add unhealthy example

### LINKED ISSUE #2601: Clarify and explain health response payload content and implications
### Description
The example payloads in our Health API and Health reference docs assume everything is working perfectly. What does a "problem" example payload look like?

Our docs also don't describe the different between an alarm and the health info. How are these different?

Clarification from Engineering: "We do not interpret the health response, meaning we do not change the response code based on the response we get from etcd.” See https://github.com/sensu/sensu-go/blob/master/backend/apid/routers/health.go#L38-L54.

Also it looks like the 500 in our Health API doc should be removed. That endpoint only returns either 200 (OK) or 400 [which indicates the timeout query parameter provided in the request is invalid (e.g. not a number)].

This information would be helpful for users who want to add monitoring for their monitoring. They need to know what Health problems could occur and what they would look like before they happen.

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/sensu-go/latest/reference/health/
https://docs.sensu.io/sensu-go/latest/api/health/

Possibly add this to the "monitor the monitor" guide being developed in https://github.com/sensu/sensu-docs/pull/2568

### Expected behavior
Users can find a "problem" example payload for the Health API that could be helpful in setting up monitoring for their monitoring.

### Current behavior
The Health payload examples in our docs currently assume everything is working perfectly

### Corrections needed
Add examples that demonstrate what a health payload looks like when there are problems in our docs

hillaryfraley:
## Description
- Replaces `500` with `400` in health API doc
- Updates healthy cluster example in health reference doc
- Adds unhealthy cluster example in health reference doc and note to explain OK HTTP response vs. unhealthy cluster status

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2601


palourde (approved):
👌

PR #2706, opened 2020-09-03
Updating the http check for /health to use the ruby check

asachs01:
## Description
Addresses the issue where regardless of what check is used, the `/health` endpoint won't return anything other than a 200.


agoddard (approved):
👍 lgtm!

Issue #2715, opened 2020-09-09
Pages should be listed in the Reference Index based on the current docs version in the URL


hillaryfraley:
### Description
To allow us to ship the observability pipeline changes (which resulted in our reference docs being spread throughout several sections of the docs rather than the single Reference directory), I created a workaround shortcode (typeListing.html) for listing all of the reference docs on a single Reference Index page.

The workaround is manually updating Line 5 in https://github.com/sensu/sensu-docs/blob/master/layouts/shortcodes/typeListing.html with the docs version that should be listed in the reference index.

This currently works because we only have one version with a reference index, 6.0. However, it will not work when we publish the 6.1 docs -- we will need the Reference Index for 6.0 to list only the 6.0 docs at the same time that the Reference Index for 6.1 lists only the 6.1 docs.

Edit: as a further workaround, I've created typeListing60.html and typeListing61.html so that I could create a working 6.1 docset for now. I can revert to a single typeListing shortcode after this is resolved.

### Category

This is an issue with:

- [x] bug (bug in site build, layout, or functionality)
- [ ] enhancement (add new or updated site functionality)

### Affected pages
https://github.com/sensu/sensu-docs/blob/master/layouts/shortcodes/typeListing.html (created to list pages with `type: "reference"` in the front matter on page https://docs.sensu.io/sensu-go/latest/reference/)

### Expected behavior
The reference index (https://docs.sensu.io/sensu-go/latest/reference/) should list all pages with `type: "reference"` in the front matter for only the current version of the docs, based on the URL. If a user changes to another docs version, the reference index page should update to list the reference docs for that version.

### Current behavior
The reference index lists pages for the docs version manually specified in line 5 of https://github.com/sensu/sensu-docs/blob/master/layouts/shortcodes/typeListing.html

### Changes needed
Replace line 5 in https://github.com/sensu/sensu-docs/blob/master/layouts/shortcodes/typeListing.html with code that determines the docs version from the current URL to produce the list of references at https://docs.sensu.io/sensu-go/latest/reference/.

### Other notes
The Reference Index page does not exist in pre-6.0 versions of the docs. For 5.x versions, the reference directory page uses the directoryListing.html shortcode.

hillaryfraley:
@rnevius I'm not sure if I explained the issue very clearly -- please let me know if it doesn't make sense.

PR #2716, opened 2020-09-10
Add 6.1 docset

hillaryfraley:
## Description
Creates 6.1 docset

## Motivation and Context
Support 6.1 release


hillaryfraley:
Includes a hacky workaround to use typeListing60 and typeListing61 shortcodes to produce the reference index in docs versions 6.0 and 6.1. I may need to remove them and replace with the correct solution when https://github.com/sensu/sensu-docs/issues/2715 is resolved.

PR #2725, opened 2020-09-11
Fix `typeListing` shortcode to work for any 6.x Reference Index

### LINKED ISSUE #2715: Pages should be listed in the Reference Index based on the current docs version in the URL

rnevius:
## Motivation and Context

Resolves #2715


hillaryfraley:
Thank you for doing this so quickly, @rnevius.

hillaryfraley (approved):
works like a charm! Thanks again.

PR #2764, opened 2020-09-30
Correct 6.0 release notes re: web UI in instances built from source

hillaryfraley:
## Description
Update note about breaking change to decouple the web UI from the backend. The note currently says this affects our binary distributions, but it should say that this affects instances built from source.

## Motivation and Context
Correct the current note, which says this affects our binary distributions.


hillaryfraley:
@jamesdphillips I believe we will also need to update the instructions at https://github.com/sensu/sensu-go/blob/master/README.md#installation, but I'm not sure what instructions to add or modify.

cwjohnston (approved):

PR #2823, opened 2020-10-30
Enhance and expand content for Deploy Sensu category index page

hillaryfraley:
## Description
Adds content to the Deploy Sensu category index page at https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/.

## Motivation and Context
Expand placeholder content for category index page so that it is more helpful to readers.

### DIFF
--- a/content/sensu-go/5.20/operations/deploy-sensu/_index.md
+++ b/content/sensu-go/5.20/operations/deploy-sensu/_index.md
@@ -5,23 +5,51 @@ product: "Sensu Go"
 version: "5.20"
 weight: 10
 layout: "single"
-toc: false
+toc: true
 menu:
   sensu-go-5.20:
     parent: operations
     identifier: deploy-sensu
 ---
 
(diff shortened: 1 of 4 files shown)


hillaryfraley:
@cwjohnston I think this is pretty much the content format we discussed (following a "story" that summarizes how readers might use the pages within the category). I added some headings to group the content by purpose (hopefully...)

If this is 👍  I can propagate to 6.0. Wdyt?

cwjohnston (approved):
Overall this looks good to me. My one reservation is that the mention of assets and other pipeline concepts feels a little disconnected. Are we mentioning it because we want to provide a path to pipeline concepts after they have installed Sensu?

hillaryfraley:
Good point -- I will remove it. I now remember we are planning to move that content into a Plugins section anyway, so we'll have to remove that text from the Deploy Sensu index page soon-ish anyway. Thank you!

PR #2867, opened 2020-11-30
Event filter is_incident allows events with unknown and custom status

### LINKED ISSUE #2865: Description of is_incident filter is inaccurate
### Description

Docs describe is_incident filter as discarding events when the status does not indicate a warning or critical status. 

### Category

This is an issue with:

- [x] errata (error in doc content)

### Affected pages

https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-filter/filters/#built-in-filter-is_incident

### Expected behavior

Explanation of `is_incident` filter behavior accurately represents handling of events with `unknown` status (status >= 3)

### Current behavior

Table and textual description indicate that events with `unknown` status will be discarded.

### Corrections needed

Table and textual description should indicate that events with check status of `unknown` and resolutions of such events will be allowed by the `is_incident` filter

### Additional context

The suggested corrections are based on my interpretation of the `IsIncident` function defined here in 6.1.3: https://github.com/sensu/sensu-go/blob/v6.1.3/api/core/v2/event.go#L97-L100 

Blame view indicates the function hasn't changed for three years, so it likely applies to all currently supported versions.

Customers have reported unexpected behavior where resolutions for events with `unknown` check status are allowed by the `is_incident` filter. I believe this is due in part to the name of the function (reasonable people may disagree as to whether `unknown` status constitutes "an incident"), but also attributable to the functionality currently described in the docs.

hillaryfraley:
## Description
Updates event filter reference doc to correct the event statuses that is_incident allows (1, 2, and 3+, as well as resolution events)

## Motivation and Context
Closes #2865


hillaryfraley:
@cwjohnston I specified "unknown or custom status" because https://github.com/sensu/sensu-go/blob/v6.1.3/api/core/v2/event.go#L97-L100 seems to say it will allow anything except status == 0. I think this is OK, but I thought I should double-check with you since the issue only mentions unknown.

PR #2873, opened 2020-12-02
Document #sensu.io.graphic appendix for displaying inline images in web UI

hillaryfraley:
## Description
- Adds note to explain when and how to use the `#sensu.io.graphic` anchor link to ensure linked images from services like Grafana will appear in the web UI in the web UI config reference doc.
- Adds link that uses `#sensu.io.graphic` to all examples in the web UI config reference and API doc.

## Motivation and Context
Appendix added in https://github.com/sensu/sensu-enterprise-go/blob/main/dashboard/src/app/util/url.ts#L22

### DIFF
--- a/content/sensu-go/6.1/api/webconfig.md
+++ b/content/sensu-go/6.1/api/webconfig.md
@@ -53,7 +53,8 @@ HTTP/1.1 200 OK
           "steamapp://34234234",
           "//google.com",
           "//*.google.com",
-          "//bob.local"
+          "//bob.local",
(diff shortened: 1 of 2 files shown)


hillaryfraley:
@jamesdphillips I can't tell from the PR whether this applies to all current versions (5.20, 5.21, 6.0, and 6.1) or only 6.1 (published since [August 21](https://github.com/sensu/sensu-enterprise-go/pull/1179)).

jamesdphillips:
I'll double check but I believe it only applies to 6.1 and later￼

hillaryfraley:
James confirmed via slack that this only applies as of 6.1.

PR #2926, opened 2021-01-05
Add note re: auto-lower-case for labels and annotations

### LINKED ISSUE #2925: Add documentation about key down-casing when using configuration files
### Description
The keys defined in `agent.yml` or `backend.yml` are down-cased when read by a third-party library we use (Viper). This is behaviour can be a source of confusion and should be documented as changing how it works is not an option at this time. See https://github.com/sensu/sensu-go/issues/4096#issue-747682348 for an example of this behaviour.

### Category

This is an issue with:

### Affected pages

I'm not sure where which page(s) should document this limitation.

hillaryfraley:
## Description
Adds notes in agent, backend, and entities references for the labels and annotations spec tables to clarify that keys for labels and annotations defined in agent.yml and backend.yml are automatically modified to all-lower-case.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2925

## Review Instructions
I took a quick spin through the docs to make sure we don't have examples showing upper-case labels or annotations to confuse people. All labels and annotations in examples use lower-case keys.

If people define labels or annotations using command line flags, can they use upper-case there? In other words, does this only affect labels and annotations defined directly in agent.yml or backend.yml?

### DIFF
--- a/content/sensu-go/5.21/reference/agent.md
+++ b/content/sensu-go/5.21/reference/agent.md
@@ -857,7 +857,9 @@ allow-list: /etc/sensu/check-allow-list.yaml{{< /code >}}
 
 | annotations|      |
 -------------|------
-description  | Non-identifying metadata to include with event data that you can access with [event filters][9] and [tokens][27]. You can use annotations to add data that is meaningful to people or external tools that interact with Sensu.<br><br>In contrast to labels, you cannot use annotations in [API response filtering][25], [sensuctl response filtering][26], or [web UI view filtering][54].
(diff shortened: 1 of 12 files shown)


cwjohnston (approved):

cwjohnston:
FWIW, I believe this change needs to be applied to all current versions of Sensu Go docs.

> If people define labels or annotations using command line flags, can they use upper-case there? In other words, does this only affect labels and annotations defined directly in agent.yml or backend.yml?

Not sure on this, I expect @amdprophet can comment.

hillaryfraley:
@amdprophet confirmed that case is maintained as specified for labels/annotations listed via command line flags. See https://sensu.slack.com/archives/C60EEQFH8/p1609869538211500

PR #3015, opened 2021-02-19
fix agent websocket default port number in backend reference

cwjohnston:
## Description

Corrects description of default port for agent websocket transport in backend reference.

## Motivation and Context

Fix a mistake I suggested in code review.


hillaryfraley (approved):
😬  I should have caught this. Thank you!

PR #3085, opened 2021-03-29
Update docs to explain wrapped-json and add examples

### LINKED ISSUE #3084: Clarify wrapped-json format and explain implications for event handling
### Description
We should provide a clearer explanation for wrapped-json format and explain how a resource retrieved from the API differs from a resource definition retrieved via sensuctl command.

Users also need to know that when Sensu sends events through the pipeline to handlers (i.e. events generated by checks), the events are in JSON format, but not wrapped-json.

Only enterprise APIs return wrapped resources; Core API returns unwrapped ones.

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/sensu-go/latest/sensuctl/#preferred-output-format
https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#event-format
https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#view-events
https://docs.sensu.io/sensu-go/latest/api/#data-format

I think this also requires correcting the command examples to retrieve resource definitions throughout the doc. If they're going to use the definition with sensuctl, the info command needs to use `--format wrapped-json`.

### Expected behavior
- The docs explain what wrapped-json means and how it is different from JSON in the Sensu context
- The docs explain that check-generated events are not wrapped -- the events sent to handlers through the Sensu observability pipeline are not in wrapped-json format
- Example sensuctl commands for getting resource definitions are correct for use with sensuctl

### Current behavior
- Docs do not explain how wrapped-json is different from JSON -- description is generally lacking
- Docs are not explicit that the check-generated events are not wrapped, so the events Sensu is sending to handlers are not what we call the "complete resource definition"
- Docs provide sensuctl examples that retrieve wrapped-json using `json` (incorrect) and do not explain that the purpose of retrieving with wrapped-json is so you can use the definitions with sensuctl.

### Corrections needed
- Improve the wrapped-json definition
- Clarify what a check-generated event looks like when Sensu sends it to a handler
- Correct all senusctl commands to use the correct format

hillaryfraley:
## Description
- Improves the wrapped-json definition in sensuctl and API index pages
- Adds an example of an event in `json` format (as Sensu would send it to a handler) in the Event reference
- Corrects sensuctl commands to include an example for wrapped-json when the output will be used for sensuctl create
- Corrects sensuctl commands to produce complete resource definitions so that they use `wrapped-json` format (since that is what we're showing in the output)

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/3084

### DIFF
--- a/content/sensu-go/6.2/api/_index.md
+++ b/content/sensu-go/6.2/api/_index.md
@@ -50,7 +50,7 @@ In terms of output formats, the [Sensu API][9] uses `json` output format for res
 For APIs that are not in the `core` group, responses are in the `wrapped-json` output format.
 The `wrapped-json` format includes an outer-level `spec` "wrapping" for resource attributes and lists the resource `type` and `api_version`.
 
-Sensu sends events to the backend [in `json` format][28], without the `spec` attribute wrapper or `type` and `api_version` attributes.
+Sensu sends events to the backend in [`json` format][28], without the `spec` attribute wrapper or `type` and `api_version` attributes.
 
 ## Versioning
 
--- a/content/sensu-go/6.2/api/_index.md
+++ b/content/sensu-go/6.2/api/_index.md
@@ -45,7 +45,11 @@ See the [RBAC reference][3] for more information about configuring Sensu users a
 ## Data format
 
 The Sensu API uses JSON-formatted requests and responses.
-In terms of [sensuctl output types][1], the Sensu API uses the `json` format, not `wrapped-json`.
+In terms of output formats, for `core`-group API responses, Sensu uses `json` format.
+For [commercial][8] APIs, Sensu uses `wrapped-json` format for API responses.
+The `wrapped-json` format includes an outer-level "wrapping" that lists the `type` and `api_version` attributes in the resource definition.
+
+Sensu sends events to the backend [in `json` format][28], without the `type` and `api_version` attributes.
 
 ## Versioning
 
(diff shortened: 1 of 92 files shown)


content/sensu-go/6.2/api/_index.md
```diff
@@ -45,7 +45,11 @@ See the [RBAC reference][3] for more information about configuring Sensu users a
 ## Data format
 
 The Sensu API uses JSON-formatted requests and responses.
-In terms of [sensuctl output types][1], the Sensu API uses the `json` format, not `wrapped-json`.
+In terms of output formats, for `core`-group API responses, Sensu uses `json` format.
+For [commercial][8] APIs, Sensu uses `wrapped-json` format for API responses.
```
majormoses: To clarify the APIs that are specific to commercial or by having a commercially build and licensed versions all calls to either sets of APIs? I think its the former based on how this reads but I wanted to double check.

majormoses (approved):
LGTM

I think there could be a little bit more explicit language to help understand where exactly `wrapped-json` applies, I am still unclear on when its needed. I don't think I have ever used `wrapped-json` directly, was that always a thing and abstractions kept me completely unaware? I have been using commercial sensu since our transition to `sensu-go` and am a bit lost on this one.

hillaryfraley:
Thank you @majormoses, you are right. I added a more explicit description of the significance of the output types in its own subsection.
edit: and clarified the core/not-core API distinction with a link to the URL format section.

echlebek (approved):

hillaryfraley:
Thanks @echlebek ! I'll propagate these changes to all versions and get this merged.

PR #3089, opened 2021-03-31
Update token substitution example

### LINKED ISSUE #3088: Token example documentation
### Description
The token example in https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-schedule/tokens/#example-token-substitution-for-check-thresholds should state a few things more explicitly:
- The example setup is an example -- you can use tokens in hooks and checks, but you don't have to always have a corresponding hook that uses tokens for every check that uses tokens.
- The example entity is not something users have to create. Instead, they should add the labels to existing entities to correspond to the keys they are using for token substitution.

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-schedule/tokens/#example-token-substitution-for-check-thresholds

hillaryfraley:
## Description
In the token reference, revises the example to demonstrate concepts more clearly: a list of steps that present the check first, then explain the command defaults, then explain how to substitute different threshold values by editing entities and adding labels. Then add the hook config example.

Also clarifies that users do not need to create a new entity -- they just edit the entities they already have for which they want to specify different thresholds from the default.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/3088


hillaryfraley:
@agoddard I confirmed that this updated workflow works, but I want to make sure that I described the commands correctly (especially the hook command).

PR #3098, opened 2021-04-05
Update css to nowrap table headers

### LINKED ISSUE #2202: Fix table content wrapping
### Description
Longer table headers break at the first column even when the other columns do not contain any header content.

Table columns do not expand to fit the longest content. Table columns with short content take up the same space as columns with more content.

I don't think we can fix this with markdown, but I want to investigate options and see what we can do.

### Category

This is an issue with:

### Affected pages
All docs pages with tables with a single header (like https://docs.sensu.io/sensu-go/latest/api/checks/#checkscheckhookshook-delete-specification)

All docs pages with content that wraps in table rows (like https://docs.sensu.io/sensu-go/latest/reference/filters/#event-attributes-available-to-filters)

### Expected behavior
Long headings extend across the full width of the table when only a single column has heading content OR break at logical points rather than randomly

Table columns expand to fit the longest content they contain.

### Current behavior
Longer content breaks randomly (including in the middle of words) even when no other columns contain content or have too much whitespace. It is unsightly.

### Changes needed
Hillary to investigate and come up with a good idea for fixing or mitigating this.

hillaryfraley:
## Description
Adds `overflow-x: visible;` and `white-space: nowrap;` to `th` in _table.scss for table header rows to prevent long table header text from wrapping to fit the width of the entire first column.

Adds a wrap override for screen size smaller than 600px so that table headings are not cut off on mobile devices.

I could not fix the way the text wraps in table cells themselves.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/2202

## More Information

Before:
before

After (screen size >= 600px):
after-desktop

After (screen size < 600px):
after-lessthan600px

### DIFF
--- a/content/sensu-go/5.21/operations/deploy-sensu/generate-certificates.md
+++ b/content/sensu-go/5.21/operations/deploy-sensu/generate-certificates.md
@@ -117,7 +117,7 @@ In issuing certificates for cluster members, the IP address or hostname used in 
 
 This guide assumes a scenario with three backend members that are reachable via a `10.0.0.x` IP address, a fully qualified name (e.g. `backend-1.example.com`), and an unqualified name (e.g. `backend-1`):
 
-Unqualified name | IP address | Fully qualified domain name (FQDN) | Additional names     |
+Unqualified<br>name | IP address | Fully qualified<br>domain name<br>(FQDN) | Additional<br>names |
 -----------------|------------|------------------------------------|----------------------|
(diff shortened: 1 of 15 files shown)


hillaryfraley:
Still testing for tables long heading text in multiple columns. (start at Control Access section)

edit: testing done. Used `<br>` as needed in tables with too many columns to allow overflow-x text.

PR #3121, opened 2021-04-19
Update redirect for ready-only to use an alias

hillaryfraley:
## Description
Add alias for /sensu-go/latest/guides/create-a-ready-only-user/ in /sensu-go/latest/operations/control-access/create-read-only-user for 6.2 doc to test using alias for redirect rather than static.json.

## Motivation and Context
Updating docs wiki, which instructs to use an alias for redirects when possible (before resorting to static.json)


hillaryfraley:
Tested -- this doesn't work. We'll have to stick with static.json redirects for now.

PR #3129, opened 2021-04-22
Add JS to clip.js to track code copy clipboard clicks

hillaryfraley:
## Description
Attempt to capture number of times docs users click to copy code examples in our docs.

Repurposes the code from https://github.com/sensu/sensu-docs/pull/3115 -- combines it with the clip.js file added in https://github.com/sensu/sensu-docs/pull/1229

## Motivation and Context
https://sensu.slack.com/archives/C6AF34L6R/p1619056103151700


hillaryfraley:
Sources used:
- https://stackoverflow.com/questions/33106189/send-ga-events-using-onclick-event
- https://developers.google.com/analytics/devguides/collection/analyticsjs/events
- https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits
- https://developers.google.com/analytics/devguides/collection/analyticsjs/writing-plugins

hillaryfraley:
@dabria @jamesdphillips I'm hoping you can take a look at this JS I hacked together (similar to the ratings PR you reviewed a few weeks ago). This one is to track each time someone uses our clipboard button to copy a code example as a Google Analytics event.

Is it OK/correct to put the code to create the GA event in a separate function as I have it currently?

I could not figure out where to put `typeof ga === "function"` in this code. Do you know if it's needed here, and if so, where I should put it?

static/js/clip.js
```diff
@@ -22,4 +23,18 @@
...
\ No newline at end of file
+})(document, Clipboard);
+
+
+/*
+* sends clicks on clipboard buttons with .copy class as Google Analytics events
+*/
+
+(function() {
+    const pageURL = document.location.pathname + document.location.search;
```
jamesdphillips:
```suggestion
    var pageURL = document.location.pathname + document.location.search;
```

I'm not super familiar with Hugo but I don't think we're using Babel or anything to transpile more modern JS syntax to ES5. So for the sake of compatibility with as many browsers as possible lets use `var` here.

static/js/clip.js
```diff
@@ -22,4 +23,18 @@
...
+
+    var button = document.getElementsByClassName("copy");
+    button.addEventListener("click", function() {
+        ga('send','event','Clipboard','Code copied',pageURL);
+    });
```
jamesdphillips:
```suggestion
    var buttonSet = document.querySelectorAll(".copy");
    buttonSet.forEach(function (btn) {
      btn.addEventListener("click", function() {
        // noop if google analytics isn't initialized
        if (typeof ga !== "function") {
          return;
        }
        ga('send','event','Clipboard','Code copied',pageURL);
       });
     });
```

I don't think this is going to work, afaik `getElementsByClassName` returns an `HTMLCollection` (a _list_ of nodes). As such we'll need to iterate over them and bind listeners to each.

jamesdphillips (changes requested):

static/js/clip.js
jamesdphillips: Updated suggestion to include guard for case where google analytics is missing.

jamesdphillips (approved):

jamesdphillips:
Obviously haven’t been able to test the code but it looks good to my eyes.

hillaryfraley:
Thank you!

PR #3140, opened 2021-05-05
Update details about PostgresHealth object in health API and reference docs

hillaryfraley:
## Description
In 6.2 health API doc:
- Adds note after backend example that payload will not included PostgresHealth if instance is not configured to use PostgreSQL
- Removes PostgresHealth object from agent transport example

In 6.2 and 6.3 health reference docs:
- Mentions agent transport endpoint in intro

In 6.3 health API doc:
- Adds note after backend and agent transport examples that payload will not included PostgresHealth if instance is not configured to use PostgreSQL

## Motivation and Context
Discussion at https://sensu.slack.com/archives/C60EEQFH8/p1620229790278900

### DIFF
--- a/content/sensu-go/6.2/api/health.md
+++ b/content/sensu-go/6.2/api/health.md
@@ -49,6 +49,10 @@ HTTP/1.1 200 OK
 }
 {{< /code >}}
 
+{{% notice note %}}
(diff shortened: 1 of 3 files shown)


fguimond (approved):
lgtm

PR #3147, opened 2021-05-07
Replace clipboard icon to copy code with copy/copied button

hillaryfraley:
## Description
Attempting to edit code and css added in https://github.com/sensu/sensu-docs/pull/1229 to create a code copy button that says "Copy" and briefly changes to "Copied!" when clicked.

## Motivation and Context
Comparing analytics on clipboard clicks vs. click tracking via crazyegg suggests that more users double/triple click to highlight code examples rather than using the clipboard copy button. We suspect we can encourage more people to use the one-click-copy function if we make its purpose more obvious and provide some indication that the copy function worked.


hillaryfraley:
Everything is working! The button to copy code starts at "Copy" and switches to "Copied!" when clicked. After 2 seconds, it switches back to "Copy" again.

hillaryfraley:
Tested in Safari and firefox as well. I think this is ready for review!

PR #3153, opened 2021-05-12
Change code click event to track pre instead of clipboard icon

hillaryfraley:
## Description
Changes the trigger for tracking clicks to copy code from `.code` to `pre` -- I believe this should track all clicks inside code examples as events. If I'm correct, we should see events in Google Analytics under the "Pre test" action.

## Motivation and Context
We were previously tracking clicks on the clipboard icon to determine the number of times code is copied. However, we discovered via crazyegg snapshot that more people seem to double-click or click-and-drag to highlight and copy from their keyboard.

Changing the trigger for a code-associated click event to `pre` should give us a more accurate picture of the number of people copying our code examples. The code to place the clipboard icon itself uses `pre`.

### DIFF
--- a/static/js/clip.js
+++ b/static/js/clip.js
@@ -40,7 +40,7 @@
         if (typeof ga !== "function") {
           return;
         }
-        ga('send','event','Pre test','Testing pre query selector',pageURL);
+        ga('send','event','Code examples','Clicks in code examples',pageURL);
        });
      });
 })();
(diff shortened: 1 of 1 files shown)


hillaryfraley:
Update: the "Pre test" clicks do appear as events in Google Analytics, so this solution seems to work.

We know this method of tracking clicks will probably result in over-counting. Essentially, a user may highlight and copy code by clicking the clipboard button (one click), clicking and dragging (one click), or triple-clicking (at least 3 clicks, but more if repeated for multiple lines of code). To mitigate this, we plan to create a crazyegg snapshot for the docs page with the most clicks to record the true number of code copies and get a better estimate of inflation among the Google Analytics click counts.

PR #3191, opened 2021-06-03
Optimize order of config options in agent.yml and backend.yml examples

### LINKED ISSUE #3166: Optimize parameter grouping in example backend.yml and agent.yml
### Description
The example backend and agent config files should group together parameters that would be the same for most installations to make it easier for users identify and focus on the parameters that need a unique configuration.

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/sensu-go/latest/files/agent.yml
https://docs.sensu.io/sensu-go/latest/files/backend.yml

### Expected behavior
In the example agent.yml and backend.yml files, the parameters that are typically the same for every installation are grouped together and separately from the parameters that usually require unique configuration. This will allow users to reuse parameters without picking through to find the parameters they likely need to configure.

### Current behavior
Parameters that are usually the same are mixed with parameters that usually need configuration, which makes it harder to use our example config files as a starting point. 

### Corrections needed
Update the example agent.yml and backend.yml files to group parameters that are typically the same together and separately from parameters that usually require unique configuration.

hillaryfraley:
## Description
Optimizes the way config options are grouped in the example agent.yml and backend.yml files so that it's easier for users to understand which options are often the same among agents/backends and which usually require customization.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/3166


content/sensu-go/6.3/files/agent.yml
```diff
@@ -1,55 +1,63 @@
...
-#password: "P@ssw0rd!"
 
 ##
-# monitoring configuration
+# often customized
 ##
+#allow-list: /etc/sensu/check-allow-list.yaml
+#backend-url:
+#  - "ws://0.0.0.0:8081"
+#  - "ws://0.0.0.0:8082"
```
echlebek: This isn't a valid URL, perhaps use `ws://example.com:8081`.

`0.0.0.0` is only a valid address as a way to tell servers to bind to all addresses in a system. Clients, like sensu-agent here, won't understand it.

Additionally, it might be helpful to include a URL like `wss://example.com:8081`, to point out that a secure websocket connection has a different scheme than a plaintext one.

echlebek (commented):
Hi, I saw your request for a review in the engineering slack :)

I'm not sure laying out the example this way makes more sense than how it was laid out before. Perhaps I do not understand the rationale?

Most example configuration files have options grouped by functionality. For instance, the postgresql example configuration file groups options this way: https://github.com/postgres/postgres/blob/master/src/backend/utils/misc/postgresql.conf.sample

By attempting to group by what we imagine to be most or least customized, we scatter the configuration options for functions around the file, making it harder to locate options by function.

hillaryfraley:
@echlebek thank you very much. I'll talk to Anthony in case I misunderstood something about the reasoning and I'll fix those wss examples (I think they may be elsewhere in the docs too... 😬 )

agoddard:
@echlebek we often see users copy and paste TLS config between backends when configuring TLS on a cluster. Presently all of the TLS related settings are spread out in the file and it leads to users missing some of them when copying and pasting. I think consolidating all of the TLS settings in one block would make life easier for people manually configuring TLS on a cluster.

echlebek:
@agoddard Oh I totally agree with having a TLS section! But I don't agree with having a section titled "often the same among agents"... make sense?

echlebek (approved):

agoddard (approved):
:tada:

PR #3210, opened 2021-06-15
Specify API port in Okta instructions

### LINKED ISSUE #3208: Possible clarifications on OIDC instructions
### Description

We had a ticket with a customer who experienced an issue with OKTA authentication - and it turns out, the port they had in the redirect URL was incorrect (`8000` instead of `8080` - `8000` is the `web port` for their configuration). Our thinking was that perhaps it might be valuable to call out that the proper port to use here is the `API port` rather than the `web port`?

### Category

This is an issue with:

### Affected pages

https://docs.sensu.io/sensu-go/latest/operations/control-access/oidc-auth/#create-an-okta-application

### Expected behavior

We're wondering if there is value in making it more explicit that the `<api_url>` should also use the `<api_port>` in step 5 of these instructions? 

### Additional context

Tl;DR customer accidentally glossed over that it should be the API port, which led to a ticket and eventually a call with the customer to resolve. It's my opinion that this is already pretty clear so if this is not changed then it's completely understandable! 

* Browser(s) and version(s): 
* OS:

hillaryfraley:
## Description
Specifies the API port to use in the API URL is 8080 in step 5 in https://docs.sensu.io/sensu-go/latest/operations/control-access/oidc-auth/#create-an-okta-application
## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/3208

### DIFF
--- a/content/sensu-go/6.0/operations/control-access/oidc-auth.md
+++ b/content/sensu-go/6.0/operations/control-access/oidc-auth.md
@@ -403,7 +403,8 @@ The steps may be different if you are using the Okta Developer Console.
 5. In the *Create OpenID Connect Integration* window:
    - *GENERAL SETTINGS* section: in the *Application name* field, enter the app name.
 You can also upload a logo in the  if desired.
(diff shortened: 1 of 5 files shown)


mgibson323 (approved):
lgtm :shipit:

PR #3229, opened 2021-06-25
Add filter example to disable alerts without silence

hillaryfraley:
## Description
Adds example filter for disabling alerts without creating silences in event filter reference.

## Motivation and Context
https://discourse.sensu.io/t/disabling-handlers/2613


hillaryfraley:
Jef approved via slack DM

PR #3249, opened 2021-07-15
6.3: Add poll_interval attribute for GlobalConfig resource

hillaryfraley:
## Description
Adds documentation for the poll_interval attribute for the GlobalConfig resource

## Motivation and Context
https://github.com/sensu/sensu-enterprise-go/pull/1576

### DIFF
--- a/content/sensu-go/6.3/api/webconfig.md
+++ b/content/sensu-go/6.3/api/webconfig.md
@@ -47,6 +47,7 @@ HTTP/1.1 200 OK
     "spec": {
       "always_show_local_cluster": false,
       "default_preferences": {
+        "poll_interval": 120000,
         "page_size": 50,
         "theme": "sensu"
       },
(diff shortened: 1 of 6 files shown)


hillaryfraley:
@jamesdphillips a few questions:

1. Is the poll_interval attribute required?
2. If it is required, what is the default value?
3. The PR https://github.com/sensu/sensu-enterprise-go/pull/1576 description made me wonder if there are different default intervals depending on the active page in the web UI. If so, which does this poll_interval attribute control (active page vs. inactive page)?

jamesdphillips:
@hillaryfraley great questions:

1. The `poll_interval` field is not required and I would argue should be avoided unless one is experiencing heavy load from clients.
2. There is no default value. If the `poll_interval` hasn't been set, the frontend will use whatever default has been configured for the view and the defaults for different views can indeed differ. For instance the context switcher, (sometimes referred to as the namespace switcher,) updates less frequently than the dashboard or events views. It's important to note that when the `poll_interval` is set that it is a _global_ preference and that _all_ views are consume it's value, to reuse the example from above, if the value is set, both the context switcher and dashboard views will use the `poll_interval` regardless of the view's default.
3. See above.

jamesdphillips (changes requested):
As per comment, the default is view dependent. Otherwise this looks great :D

PR #3270, opened 2021-07-29
Document potential issue with using only default storage after provisioning attached storage

hillaryfraley:
## Description
Document the performance reduction associated with provisioning attached storage but using only default storage as a potential reason for the "took too long to execute" backend log message in https://docs.sensu.io/sensu-go/latest/operations/maintain-sensu/troubleshoot/#symptoms-of-poor-performance

## Motivation and Context
https://sumologic.slack.com/archives/C024XK35Z3Q/p1627579140121600


echlebek (commented):
We can make a diagnostic suggestion:

Ensure that the attached storage is formatted and mounted. Typically it would be /dev/nvme1n1, and _not_ /dev/nvme0n1.

echlebek (commented):
Also note that this is 100% specific to AWS, so make sure that this is not offered as general advice.

PR #3279, opened 2021-08-10
New docs.sensu.io landing page

### LINKED ISSUE #2915: New docs.sensu.io landing page
### Description

As Sensu Core and Enterprise reached EOL, we changed the text and links on the tiles at our landing page, docs.sensu.io (/layouts/_index.html), but haven't changed the page itself. We should make better use of the docs landing page to help people find Sensu Go docs and represent the full range of documentation we provide.

### Category

This is an issue with:

### Affected pages
https://docs.sensu.io/

### Changes requested
- Overhaul `/layouts/index.html` to reproduce to the [wireframe design](https://github.com/sensu/sensu-docs/files/5718352/12-15-20_docs.sensu.io_wireframe.pdf), using new partial templates where appropriate to reduce clutter and repetitive HTML.
- Each element on the landing page should have a mouseover effect of some kind to help users navigate and orient themselves on the page.
- We should be able to link each element on the landing page to a corresponding page in the docs.
   - Here is a [map of the current link plan](https://github.com/sensu/sensu-docs/files/5851995/Wireframe-to-docs.mapping.xlsx) in case you need it for any reason, but please use link placeholders in your PR -- I can add the links when I'm testing the PR.
   - In one case (the blue "Need help with a question?" element), we will need to link the text in the element rather than the element itself because the element includes two different links.
   - In another case (the green "New to Sensu?" element), we will need to link to a section on the landing page itself (in other words, we want to link the "New to Sensu?" element to the "Learn Sensu" section at the bottom of the page).
- Where the wireframe shows the simplified observability pipeline image, use this image instead:
observability-pipeline
   - We plan to replace the PNG image with the SVG inline code for the same image. We're going to use our existing image for now while we think through ideas for a redesigned version.
- Logos for each integration should appear more or less evenly in the "Sensu Integrations" section.
   - I'm not sure how you will approach this --  here's a [zip folder of the logos](https://github.com/sensu/sensu-docs/files/5851944/logos.zip) in case you do something like use CSS to resize them, but placeholders are also fine.
   - If needed, I can alter the logo dimensions to fit your design -- just give me the specs.
- For ongoing updates and maintenance, we should be able to change the text and links on the page, add and remove elements as we add new features in Sensu, and move elements and sections to different locations on the page if needed.

### Additional context

The improved docs landing page:
- Gives users a fuller picture of the information they will find in the Sensu docs: communicates the breadth and depth of the Sensu documentation
- Provides an at-a-glance view of Sensu Go features and concepts
- Includes sections that respond to every level of familiarity with Sensu, so brand new users and experienced operators alike will find information appropriate for their needs
- Links directly to individual docs pages and offers lots of entry points, both within the docs and to other Sensu resources like Discourse

rnevius:
## Motivation and Context

Resolves #2915

## Review Instructions

Needs copy review, image replacements, and links to content.


hillaryfraley (changes requested):
@rnevius Thank you, this is great! A few requests...

1. [x] For all elements that have a link, can you add a "grow" or "shadow" hover effect (in addition to the reduced opacity) to make the mouseover more noticeable?
2. [x] I added the Sensu logo to the top left of the page. Is it possible to make the three boxes on the right a little smaller, so they are less obtrusive? And add a bit of space between to push the logo further to the left and the three boxes further to the right?
3. [x] I noticed that when I switch to a mobile view of the page, the text wraps mid-word in the three boxes at the top right. Is it possible to responsively reduce the text size in those boxes so the wrap is at spaces between words instead of in the middle of words?
4. [x] I want to add links to each of the three steps in the parallelograms (different links for each step) -- I couldn't figure out how to do that in the current configuration.
5. [x] "New to Sensu" and "Sensu Core User" boxes: rather than only linking the text, the boxes themselves should be clickable (like the purple Plugins box a bit lower on the page).
6. [x] Just a note that I replaced the observability pipeline .png image with code for an SVG that contains links.
7. [x] In the purple Plugins box, the text is underlined -- can you remove the underlining from the text (here and in other boxes that have links, like the boxes in the "Sensu at enterprise scale" and "Guides and examples" sections)?
8. [x] Can we standardize the size of the integrations logos using CSS? Or do I need to resize the image files themselves?
9. [x] Since we might want to add more logos or other linked elements within each section, can the bottom row of linked elements always be centered? Currently it seems everything left-aligns if there are fewer elements in the last row than in previous rows.
10. [x] Green sensuctl, web UI, and API boxes: can these have the same appearance and layout as the boxes in the "Sensu at enterprise scale" and "Guides and examples" sections?

rnevius:
@hillaryfraley the size of the logos will need to be standardized via image dimension/size/bounding box. The CSS is only aware of the image bounds/dimensions and not the image content.

hillaryfraley:
@rnevius, many thanks! 🎉

hillaryfraley (changes requested):
@rnevius one more request: when I preview this on a smaller device like my phone, the longer words are wrapped in the purple, blue, and white boxes under "Sensu at enterprise scale," "Guides and examples," and "Learn Sensu." Can you responsively reduce the text size in these boxes so that words don't wrap? If that isn't possible, can you force hyphenation as needed for these longer words? 

- [x] In the linked boxes under Sensu at enterprise scale," "Guides and examples," and "Learn Sensu," responsively reduce the text size in those boxes so the wrap is at spaces between words instead of in the middle of words or force hyphenation for display on smaller screen sizes

hillaryfraley (approved):
@rnevius it's perfect. Thank you so much!

PR #3470, opened 2021-11-08
Replace nagiosfoundation check with Sensu Processes Check to monitor NGINX

hillaryfraley:
## Description
In the Monitor server resources guide, replaces the nagiosfoundation check with a check from the Sensu Processes Check asset to monitor a webserver (NGINX).

## Motivation and Context
Outcome of discussion w/DAs last week


jspaleta:
Only small thing,  when creating the sensu resources from yaml/json  leave out the namespace so sensuctl can populate it just in case the user is using a non-default namespace in their sensuctl configuration.

PR #3516, opened 2021-11-29
Clarify language around upgrading Sensu

asachs01:
As part of a request in HS-27590, there needs to be some additional clarity in this document when it comes to performing a rolling upgrade. As it stands, the doc can be interpreted as if an operator should proceed through all steps first on a single backend, then upgrade the rest of the cluster members. The note I've added is an attempt to ensure that users know they should upgrade all backends before upgrading the database.


hillaryfraley (approved):
👍  I'll propagate to all versions and merge!

Issue #1223, opened 2019-02-20
Add yaml block validator to CI


rgeniesse:
## This is an issue with:

- [ ] Bug (site functionality or styling)
- [ ] Errata (fix needed for doc content)
- [ ] New content (guide wanted)
- [ ] Update (Add missing or refresh existing content)
- [X] Enhancement (add new site functionality)

## Current Behavior
Currently there is a `json` block validator that runs with CI. To go along with the eventual move to `yaml` examples with Sensu Go, a `yaml` validator should be written that does the same as the current `json` one. Language does not matter.

## Possible Solution
Write it!

hillaryfraley:
https://github.com/oxsecurity/megalinter

PR #3131, opened 2021-04-26
Add yaml block validator

hillaryfraley:
## Description
Validates YAML within tags `{{< highlight yml >}} {{< /highlight >}}` using Python's YAML library.

PR #3627, opened 2022-01-26
Updating tests for yaml linting and also changing to github actions

hillaryfraley:
I know this isn't necessarily a best practice, but this will address #1223 and also upgrades the docs to use github actions. The benefit being that we don't have to rely on Travis and the yamllinting bits are native in Github actions.

Replaces https://github.com/sensu/sensu-docs/pull/3624


hillaryfraley:
2/17/22 docs inbox cleanup discussion: we will need a Python script similar to https://github.com/sensu/sensu-docs/blob/main/tests/json-validator.py to handle yaml validation. The approach with yaml lint seems like it won't work because we only want to lint the yaml within yaml-specific code tags.

hillaryfraley (commented):
Here's a PR where I tried to duplicate Richie's Python code for JSON linting:

https://github.com/sensu/sensu-docs/pull/3131/files

It didn't work, but maybe I did something obviously wrong...?

hillaryfraley:
Today I found out about https://github.com/earldouglas/codedown -- maybe this could help?

hillaryfraley:
https://github.com/oxsecurity/megalinter

asachs01:
I think I've got it. I need a review though.

PR #3708, opened 2022-03-02
Hide copy button for code text format

### LINKED ISSUE #3003: Optionally disable one-click copy function for code-formatted text
### Description
We need to present certain text in code format without including the one-click copy function. For example, many of our docs include sensuctl responses, which are not intended to be copied. Including the copy function for this text may be confusing.

### Category

This is an issue with:

### Affected pages
All pages -- this formatting upgrade should be available to use on any page in the docs

### Expected behavior
Docs authors should be able to disable the one-click copy function so we can format text in code syntax without indicating that users should copy it.

### Current behavior
All text formatted in code syntax includes the one-click copy function.

### Changes needed
Update code that controls the one-click copy function to include a way to optionally disable to copy function as needed. I think the functionality was added in https://github.com/sensu/sensu-docs/pull/1229.

### Additional context
See https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/scale-event-storage/#configure-sensu -- the backend log example is just intended to demonstrate what the log entries will look like. It should not include the one-click copy function.

hillaryfraley:
Signed-off-by: Hillary Fraley <hfraley@sumologic.com>

## Description
Hides or does not display the code copy button for response examples.

My thoughts about how to do this:
- Hugo renders content placed inside `{{< code >}}` tags with [Chroma](https://github.com/alecthomas/chroma) for syntax highlighting.
     - If you inspect the HTML, all code examples are inside `<pre>` HTML tags, with a nested `<code>` tag.
     - The `<code>` tag includes a class based on the language specified in the Hugo `{{< code >}}` tags. For example, text inside `{{< code shell >}}` tags is rendered in HTML like `<code class="language-shell" data-lang="shell">`.
- Since all content inside code formatting tags in Hugo is rendered with the language-based class and data-lang attribute, I should be able to target this language-based information by adding some kind of if-else statement in the existing clip.js file.
- If I use `{{< code text >}}` tags for responses, the code formatting will still display properly for things like tables copied from terminal. This will produce the HTML `<code class="language-text" data-lang="text">`.
- I think I can target either the class or the data-lang attribute to use either `button.style.visibility="hidden"` or `button.style.display="none"` to avoid showing the copy button for responses (in other words, the content placed inside `{{< code text >}}` tags).

So far, this has not worked. Maybe the problem is one of these things:
- I need to place the code to remove the button after the button is available in the DOM. I think the latest commit achieves this, but I'm not sure.
- The code to hide the button is working, but it's being applied too broadly. The latest commit shows the copy button on the first code example, but then does not show it on any others. I think this means I am targeting incorrectly or incompletely. Or maybe I need to add something more than the "if" statement for hiding the button -- something that says "this only applies to the current content inside the targeted tags, so run this evaluation again every time you hit a code tag." I can't figure out how to test this.

## Motivation and Context
Closes https://github.com/sensu/sensu-docs/issues/3003


hillaryfraley:
to try next:
https://stackoverflow.com/questions/8685107/hiding-a-button-in-javascript
https://stackoverflow.com/questions/32146962/how-to-hide-button-in-javascript

hillaryfraley (commented):
I don't think the special shortcode for responses will work without a lot of CSS. We have many responses in shell tables and I can't get the tables to display properly.

hillaryfraley (commented):
In docs inbox cleanup today, we also tried this in the first function:

```

/*
 * clipboard for hugo
 */

(function(document, Clipboard) {

    var $codes = document.querySelectorAll('pre');
    //*const codelangtext = document.querySelectorAll('.language-text');

    function addCopy(element) {
        var copy = document.createElement("button");
        copy.className = "copy";
        copy.innerText = "Copy";
        element.append(copy);
    }

    if element.classList.contains('language-text'); {
        button.style.visibility="hidden";
    }
    else {
        button.style.visibility="visible";
    }
    
    for (var i = 0, len = $codes.length; i < len; i++) {
        addCopy($codes[i]);
    }

    var clipboard = new ClipboardJS('.copy', {
        target: function(trigger) {
            return trigger.previousElementSibling;
        }
    });

    clipboard.on('success', function(e) {
        e.trigger.textContent = 'Copied!';
        setTimeout(function () {
            e.trigger.textContent = "Copy";
            }, 2000);
    });

})(document, Clipboard);
```

This hides the Copy button in all code examples throughout the docs.

static/js/clip.js
```diff
@@ -6,12 +6,16 @@
...
 
     var $codes = document.querySelectorAll('pre');
+    const codelangtext = document.querySelectorAll('.language-text');
 
     function addCopy(element) {
         var copy = document.createElement("button");
         copy.className = "copy";
         copy.innerText = "Copy";
         element.append(copy);
+        if (codelangtext = true) {
```
jamesdphillips: Imperative languages are a bit confusing in this regard but a single `=` doesn't denote equity as it does in mathematics but instead _assigns_ a value to a variable. In this case this statement assigns the value `true` to the `codelangtext` variable. Essentially this means when evaluated it is equivalent to:

```ecmascript
if (true) {
  button.style.visibility="hidden";
}
```

static/js/clip.js
```diff
@@ -6,12 +6,16 @@
 (function(document, Clipboard) {
 
     var $codes = document.querySelectorAll('pre');
+    const codelangtext = document.querySelectorAll('.language-text');
 
     function addCopy(element) {
```
jamesdphillips: I think you're on the right path with this. What I'd suggest is that instead of searching for the text elements before we add the button, we instead search for the `.language-text` class within the selected pre tags, and then if we get an hits simply don't add the copy button. (This pattern is often called a guard statement, as we're _guarding_ the `addCopy` function from proceeding when one or more conditions are met.)

```suggestion
    function addCopy(element) {
      const codelangtext = element.querySelectorAll('.language-text');
      if (codelangtext.length > 0) {
        return;
      }
```

static/js/clip.js
```diff
@@ -6,12 +6,18 @@
...
+      if (codelangtext.length > 0) {
+        return;
+      }
         var copy = document.createElement("button");
         copy.className = "copy";
         copy.innerText = "Copy";
         element.append(copy);
+        if (codelangtext = true) {
+            button.style.visibility="hidden";
+        }
```
hillaryfraley:
```suggestion
```

PR #4018, opened 2022-09-19
Add refresh tokens info in OIDC reference

hillaryfraley:
## Description
Adds brief section with instructions to enable refresh tokens for OIDC providers in OIDC reference doc

## Motivation and Context
https://sumologic.slack.com/archives/C024PCF29KR/p1662659371570259

review note: I will propagate the refresh token info to all docs versions after all revision is incorporated


asachs01 (approved):
Perfect! 👍

PR #4027, opened 2022-09-23
Update Hugo version

hillaryfraley:
## Description
Updates the Hugo version to 0.101.0.

The latest version is 0.103.1, but 0.101.0 is the latest version that will build. I think it's something to do with the [changes in 0.102.0](https://github.com/gohugoio/hugo/releases/tag/v0.102.0).


hillaryfraley (commented):
I confirmed that the site builds locally and that the offline docs generate as expected with version 0.101.0. I also tested the Heroku preview, and all looks good.

hillaryfraley (commented):
The `.hugo_build.lock` file was introduced in [Hugo 0.89.0](https://github.com/gohugoio/hugo/releases/tag/v0.89.0)

PR #4039, opened 2022-09-30
Upgrade yarn dependencies

hillaryfraley:
## Description
Ran `yarn upgrade` and `yarn add node-sass` to upgrade the node-sass version and allow dependency upgrades.


hillaryfraley (commented):
Tested a locally running version and the Heroku preview. All good.

PR #4141, opened 2022-11-21
Updating store

asachs01:
It's not necessary (and may even be confusing) to include 2 different store references. I'm removing the first one.


hillaryfraley (approved):
Fixed one more instance and propagated to all docs versions

hillaryfraley:
Thank you for finding this!

PR #4190, opened 2023-05-01
Updating datastore doc to include upgrade instructions

asachs01:
## Description

Adds links to Etcd & Postgres upgrade documentation.

## Motivation and Context

Customers have asked about upgrading and noted that we didn't include those instructions. However, as we're not the maintainers responsible for Etcd/Postgres, it's best that we refer customers directly to the source when it comes to performing upgrades.


jhenderson-pro (approved):
lgtm!

